PR #5, opened 2016-07-25
Chenge the way Vulkan API version is set

Darktori:
The previous way is deprecated.

### DIFF
--- a/Quake/gl_vidsdl.c
+++ b/Quake/gl_vidsdl.c
@@ -586,7 +586,7 @@ static void GL_InitInstance( void )
 	application_info.applicationVersion = 1;
 	application_info.pEngineName = "vkQuake";
 	application_info.engineVersion = 1;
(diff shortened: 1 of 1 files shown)


Novum:
Why?

Novum:
Oh I see, deprecation. Sure, but why 1.0.6?

Darktori:
Literally the first example I found on the web (the original version did
not compile for me). I can change it to a different one.

wt., 26.07.2016, 00:40 użytkownik Axel Gneiting <notifications@github.com>
napisał:

Novum:
I don't see this being deprecated in the header. VK_API_VERSION is deprecated, not VK_API_VERSION_1_0.

Novum:
VK_API_VERSION_1_0 is also used by the official samples. Please use a later SDK version if this doesn't compile.

Issue #7, opened 2016-07-26
quakespasm: isl.c:1516: isl_surf_get_image_offset_el: Assertion `level < surf->levels' failed.


iamgreaser:
This happens when I try to start a game or run a demo.

Running very recent Mesa git i965 drivers on Skylake on Linux (git-126bd15 according to glxinfo). These drivers are compiled with `--enable-debug`.

Vulkan does work in general.

Using the registered version of the files as available on Steam.

Attempting to build a proper debug build results in an error during compilation so I may have to update my headers at some stage.

Relevant backtrace:

    (gdb) bt
    #0  0x00007ffff6554207 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
    #1  0x00007ffff655568a in __GI_abort () at abort.c:89
    #2  0x00007ffff654d1ed in __assert_fail_base (fmt=0x7ffff6689390 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=assertion@entry=0x7fffe02e73c6 "level < surf->levels", 
        file=file@entry=0x7fffe02e7240 "isl.c", line=line@entry=1516, function=function@entry=0x7fffe02e7b50 <__PRETTY_FUNCTION__.3574> "isl_surf_get_image_offset_el") at assert.c:92
    #3  0x00007ffff654d2a2 in __GI___assert_fail (assertion=assertion@entry=0x7fffe02e73c6 "level < surf->levels", file=file@entry=0x7fffe02e7240 "isl.c", line=line@entry=1516, 
        function=function@entry=0x7fffe02e7b50 <__PRETTY_FUNCTION__.3574> "isl_surf_get_image_offset_el") at assert.c:101
    #4  0x00007fffe016533d in isl_surf_get_image_offset_el (surf=surf@entry=0x20cbba0, level=7, logical_array_layer=<optimized out>, logical_z_offset_px=logical_z_offset_px@entry=0, 
        x_offset_el=x_offset_el@entry=0x7fffffffa468, y_offset_el=y_offset_el@entry=0x7fffffffa46c) at isl.c:1516
    #5  0x00007fffdea20724 in meta_copy_buffer_to_image (cmd_buffer=0x1edac00, buffer=0x1df85c0, image=0x20cbb50, regionCount=<optimized out>, pRegions=<optimized out>, forward=true) at anv_meta_copy.c:189
[...]

PR #19, opened 2016-07-28
Fix mip map generation (for rectangular textures)

SaschaWillems:
The initial staging code would generate one more copy region than number of mip levels specified. This lead to a validation error with a potential crash on some implementations due to the additional copy.

Code could look a bit nicer, but tested this to work on two different implementations (and verified with Renderdoc) and validation errors are also gone.

Refs #7


Novum:
Thanks

PR #26, opened 2016-07-30
fixed 'Buffer is accessed out of bounds', reported by cppcheck.

stevenhoving:
I wonder if this is a correct fix. I could also have fixed it by changing:
`vec3_t			lightcolor;`
to
`vec4_t			lightcolor;`

So what do you guys prefer?


Novum:
Fixed by bf274105f990707fb373889364d9bc5a322eab17

PR #41, opened 2016-08-02
Simplify VS usage

bviktor:
Use a single folder for pak files and copy this to the output folder
Document the usage changes
Add placeholder id1 folders to help users (and to prevent xcopy errors)
Extend gitignore according to the changes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ Quake/vkquake
 Quake/vkQuake.exe
 Quake/vkQuake.res
 Windows/VisualStudio/Build-vkQuake/
+Windows/VisualStudio/history.txt
+Windows/VisualStudio/id1/
--- a/Windows/VisualStudio/vkquake.vcxproj
+++ b/Windows/VisualStudio/vkquake.vcxproj
@@ -118,7 +118,8 @@
     </Link>
     <PostBuildEvent>
       <Command>copy "$(SolutionDir)\..\codecs\x86\*.dll" "$(TargetDir)"
-copy "$(SolutionDir)\..\SDL2\lib\*.dll" "$(TargetDir)"</Command>
+copy "$(SolutionDir)\..\SDL2\lib\*.dll" "$(TargetDir)"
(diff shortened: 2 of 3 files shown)


Novum:
Declined. Set the "Working Directory" under "Debugging" in "Project Settings" to your Quake installation. No need to copy around anything.

bviktor:
Yes, it's already set so VS works, the reason behind this is the other way around. This copy is there for the same reason you copy DLLs to $TargetDir (which you could also avoid by simply setting PATH under project settings).

I assumed that DLL copy is in place to allow running vkQuake.exe directly from Explorer/CMD. This copy is there for the exact same reason. Without that, you can only run vkQuake from VS or copy files around manually.

BTW DLL files are also copied in the MinGW makefile (I opened that PR too).

So I guess that for the sake of consistency, it'd only make sense to copy both pak+dll or neither. Could you please clear this up?

bviktor:
The other rationale behind this copy is the fact that copying 3 more files on top of 10-something other files is considerably less complicated than instructing users to copy pak files to `Windows\VisualStudio\id1` **or** `Windows\VisualStudio\Build-vkQuake\x64\Release\id1` **or** `Windows\VisualStudio\Build-vkQuake\x64\Debug\id1` **or** `Windows\VisualStudio\Build-vkQuake\x86\Release\id1` **or** `Windows\VisualStudio\Build-vkQuake\x86\Debug\id1`, depending on the use case.

Novum:
I inherited this from QuakeSpasm and I don't think it's a good idea to copy the DLLs either :)

bviktor:
How do you do a Windows release then?

Novum:
Manually ¯\_(ツ)_/¯

Novum:
Sorry, I don't want to be a dick about this. I'll look at it again when I'm at home and make better suggestions.

Novum:
I still think users should just set the working directory and copy the required DLLs into the Quake directory.

PR #43, opened 2016-08-04
GetSwapchainImagesKHR: error if image_count is less than requested

groleo:
On Linux with mesa-vulkan, the image_count is 4, which is greater than the 2 requested


Novum:
Yeah, that won't work like that.

Novum:
There are other places that need to be changed to fix that. I'll leave this open in the meantime.

Novum:
Fixed with f52f09eb1f1315c4fa31a522c8d4a110900decd2

PR #47, opened 2016-08-05
replace #ifdef _DEBUG with DEBUG

groleo:
I assume that the intention of using `#ifdef _DEBUG` is to enable
validation/checks only in debug builds.

But Quake/Makefile defines DEBUG, not _DEBUG.


groleo:
uhm, it might be the other way around, considering the vkquake.vcxproj defines _DEBUG

      <PreprocessorDefinitions>WIN32;_DEBUG

PR #51, opened 2016-08-07
Linux shader compilation script, uses find to compile all shaders in …

vrodic:
…directory


Novum:
Please provide the shader compilation script as a separate pull request, I will merge that.

PR #65, opened 2016-09-11
misc build updates

sezero:
Misc build updates, and 3d-party libs sync from quakespasm.


sezero:
Builds for w32 and w64 on my linux using gcc-4.5 and even gcc-3.4. Builds tested to run on a win64 with amd drivers.

PR #68, opened 2016-09-12
small stuff, sync with qs

sezero:
- remove rest of the SDL1 refs from source
- remove unnecessary _WIN32 checks from windows source
- warning fixes for bintoc.c
- sync with quakespasm-0.92.0: (missed parts in SCR_ModalMessage() and R_BuildLightMap()
- sync with quakespasm tree: rest of the small stuff, for sake of symmetry

### DIFF
--- a/Quake/Makefile
+++ b/Quake/Makefile
@@ -70,7 +70,6 @@ define do_strip
 	$(call cmd_strip,$(1));
 endef
 endif
-CFLAGS += -std=gnu99
 CFLAGS += -I$(VULKAN_INCLUDE_DIR) -DLINUX -DVK_USE_PLATFORM_XCB_KHR
 
 ifeq ($(DO_USERDIRS),1)
--- a/Quake/Makefile
+++ b/Quake/Makefile
@@ -249,11 +249,11 @@ GLOBJS = \
 	image.o \
 	gl_texmgr.o \
 	gl_mesh.o \
+	gl_heap.o \
 	r_sprite.o \
 	r_alias.o \
 	r_brush.o \
-	gl_model.o \
-	gl_heap.o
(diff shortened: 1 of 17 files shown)


Quake/Makefile
```diff
@@ -70,7 +70,6 @@ define do_strip
 	$(call cmd_strip,$(1));
 endef
 endif
-CFLAGS += -std=gnu99
```
ProfessorKaos64: Seems to compile fine on Arch Linux. What distro did you test this change on?

sezero: On 9/13/16, ProfessorKaos64 <notifications@github.com> wrote:
>> @@ -70,7 +70,6 @@ define do_strip
>>  	$(call cmd_strip,$(1));
>>  endef
>>  endif
>> -CFLAGS += -std=gnu99
>
> Seems to compile fine on Arch Linux. What distro did you test this change
> on?

Fedora 20.

Novum: Are you sure this is fine? There are still plenty of functions where I declare variables not at the beginning of the scope?

ProfessorKaos64: [I haven't seen C89 added to a project in a long while](https://i.imgflip.com/wppxk.jpg)

sezero: On 9/13/16, Axel Gneiting <notifications@github.com> wrote:
>> -CFLAGS += -std=gnu99
>
> Are you sure this is fine? There are still plenty of functions where I
> declare variables not at the beginning of the scope?

Yes, however without an explicit -ansi or similar, gcc extensions
are in effect, so unless someone attempts using gcc2 (which btw
does not support -std=gnu99), you are fine.

PR #83, opened 2016-12-07
remove some C99'isms (for loop initial declarations) so that it build…

sezero:
…s with some older

compilers. also change a few sources to have unix line endings uniformly like the rest.
tested to build with mingw (gcc-3.4.5) and mingw-w64 (gcc-4.5.4).


Novum:
Hmm lots of the diff looks like only whitespace differences?

sezero:
Splitting it.

mdeguzis:
@Novum , If I recall right, you removed these and then they were put back. I'll try to dig up the commit to see why, but it was still decided to keep them.

mdeguzis:
See: 14fa407

sezero:
> See: 14fa407

IIRC, that flag was put back in just in case, and it used to compile without it just fine;
plus, the mingw makefiles never had that flag (and it is nice to be compile without it.)
The offending cases I address in #85 seem to have been added more recently.

PR #85, opened 2016-12-07
remove some C99'isms (for loop initial declarations)

sezero:
so that it builds with some older compilers.
tested build with mingw (gcc-3.4.5) and mingw-w64 (gcc-4.5.4).

### DIFF
--- a/Quake/gl_rmisc.c
+++ b/Quake/gl_rmisc.c
@@ -1118,9 +1118,10 @@ R_CreatePipelines
 */
 void R_CreatePipelines()
 {
-	Con_Printf("Creating pipelines\n");
-
+	int render_pass;
 	VkResult err;
+
+	Con_Printf("Creating pipelines\n");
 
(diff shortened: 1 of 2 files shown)


mdeguzis:
I sure hope this doesn't break the Linux builds. This was removed then readded back. If I must, I will `+=` the flag in my package rules. See: 14fa407

sezero:
@mdeguzis: The intention is to be able to build w/o needing and extra -std flag,
ad the changeset can't ever possibly break linux builds.

mdeguzis:
Yes, I understand, as my edit noted, I misread your changes. Sorry.

Novum:
I appreciate the fixes for this. I just can't get my head to not do this when I write code :)

Novum:
Although I wonder what systems have Vulkan support and don't support C99?

sezero:
> Although I wonder what systems have Vulkan drivers but not a C99 compiler?

Probably not much.  However a cross-build environment might not have the newest
and greatest of compilers but still can produce a target binary just fine (such as the
mingw32 one I had tested.)

PR #94, opened 2017-02-18
Clear VkWriteDescriptorSet screen_warp_writes to zero

ardacoskunses:
VkWriteDescriptorSet.pNext member stay dirty causing invalid
memory access


Novum:
Thanks

PR #99, opened 2017-05-13
Swap chain composite alpha selection based on surface capabilities

SaschaWillems:
VK_COMPOSITE_ALPHA_OPAQUE_BIT_KHR [isn't supported by all devices](http://vulkan.gpuinfo.org/displayreport.php?id=1384#surface) . This change will pick the composite alpha format based on the surface capabilities.


Novum:
Can you only check for OPAQUE and INHERIT? I'm afraid the premultiplied or postmultiplied would potentially just display black with alpha = 0

PR #116, opened 2017-09-11
Use xdg config directory on Unix-like systems

chetgurevitch:
Per the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html). Basically moves the vkquake directory to XDG_CONFIG_HOME (~/.config by default) but leaves the file structure fully intact, just like gzdoom on *nix. Also uses $HOME from the environment over pw_dir.


Novum:
I think it would be better to get this merged into QuakeSpasm as it is the upstream for all non graphics related changes. What do you think?

chetgurevitch:
Yeah you're probably right, I'll see if upstream's interested. Side note managed to pick the exact time sourceforge went down to follow up on this lol.

PR #129, opened 2017-12-20
Fix LDFLAGS

kingtiger01:
LDFLAGS being replaced rather than ammended, prevents additional libraries from being specified.
EG:"-lopusfile -lFLAC -lmikmod"


Novum:
Sorry, I just saw this. Thank you

PR #165, opened 2019-03-02
Allow reallocating dynamic vertex, index and uniform buffers (fixes #164)

### LINKED ISSUE #164: Out of dynamic vertex buffer space
I'm playing the "Scourge of Armagon" mission pack, and encountered an error on the map "Research Facility" (HIP1M4).

QUAKE ERROR: Out of dynamic vertex buffer space, increase DYNAMIC_VERTEX_BUFFER_SIZE_KB

This occurs when I pick up the quad damage in the map. Its reproducable everytime.

rg3:
I'm a complete Vulkan and computer graphics newbie and this is the best thing I could come up with in a few hours, but it seems to work here with no validation errors. Let me know what you think and do not hesitate to ask for changes or maybe a different strategy. The commit message follows below:

Vertex, index and uniform buffers had a fixed size. With this change,
they can be reallocated when needed and grow to hold more data.

Because Vulkan is async, the existing buffers, their memory and
descriptor sets cannot be deleted or updated right away after allocating
new ones. Instead, the existing buffers need to be kept around for some
time until the device is done using them. That's the purpose of the
"to_be_deleted" fields added to some structures.

Dynamic buffer sizes grow with a factor of 1.5 in each reallocation, as
is common practice[1].

In addition, the common code in R_InitDynamicVertexBuffers,
R_InitDynamicIndexBuffers and R_InitDynamicUniformBuffers has been
abstracted to a single function, and all of them can be called again
when the buffers need to grow in size.

[1] https://stackoverflow.com/questions/2269063/buffer-growth-strategy

### DIFF
--- a/Quake/gl_rmisc.c
+++ b/Quake/gl_rmisc.c
@@ -85,29 +85,59 @@ static int				current_staging_buffer = 0;
 Dynamic vertex/index & uniform buffer
 ================
 */
-#define DYNAMIC_VERTEX_BUFFER_SIZE_KB	2048
-#define DYNAMIC_INDEX_BUFFER_SIZE_KB	4096
-#define DYNAMIC_UNIFORM_BUFFER_SIZE_KB	1024
 #define NUM_DYNAMIC_BUFFERS				2
 #define MAX_UNIFORM_ALLOC				2048
+
+size_t dynamic_vertex_buffer_size_kb  = 2048;
+size_t dynamic_index_buffer_size_kb   = 4096;
+size_t dynamic_uniform_buffer_size_kb = 1024;
 
 typedef struct
 {
 	VkBuffer			buffer;
 	uint32_t			current_offset;
 	unsigned char *		data;
+	size_t				size;
 } dynbuffer_t;
 
-static VkDeviceMemory	dyn_vertex_buffer_memory;
-static VkDeviceMemory	dyn_index_buffer_memory;
-static VkDeviceMemory	dyn_uniform_buffer_memory;
-static dynbuffer_t		dyn_vertex_buffers[NUM_DYNAMIC_BUFFERS];
-static dynbuffer_t		dyn_index_buffers[NUM_DYNAMIC_BUFFERS];
-static dynbuffer_t		dyn_uniform_buffers[NUM_DYNAMIC_BUFFERS];
-static int				current_dyn_buffer_index = 0;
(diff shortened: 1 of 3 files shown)


Novum:
Hey, I'm sorry, I completely missed those pull requests. I will review this in a bit.

Novum:
I feel like this is overly complex with the double buffering etc. I would just have done a wait for device idle and recreated the buffers afterwards. Shouldn't happen often.

rg3:
Thanks for taking the time to review this!

I tried that simple approach first and the problem is that by the time you realize you need to reallocate buffers because you ran out of space, you have already recorded commands using those buffers, so it's not as simple as waiting for the device to be idle before reallocating. I got validation errors like the following ones:

```
UNASSIGNED-CoreValidation-DrawState-InvalidBuffer(ERROR / SPEC): msgNum: 0 - Cannot submit cmd buffer using deleted buffer 0xb8.
    Objects: 1
        [0] 0xb8, type: 9, name: NULL
```

Commands submitted before running out of space need to be re-recorded, and that requires, I believe, even more complex logic. I thought letting those buffers stay alive for a bit longer was an easy way out of the problem.

Then again, I'm not an expert in Vulkan and not very familiar with the Quake engine.

rg3:
@Novum friendly ping. Did you have the chance to review my comment above?

By the way, when I said "Commands submitted before running out of space" I meant to say "Commands **recorded** before running out of space".

Novum:
We are super busy with Doom Eternal, I will try to review it again.

Novum:
So one of the issues I found is that you are just increasing the size by 1.5x and then you just assume that the requested allocation will fit in that, but there is nothing preventing it from being e.g. 2x the old buffer size.

I also dislike this `memory_buffer_pair_t` thing. I just would add global double buffered VkBuffer & vkDescriptorSet garbage lists. Then just delete the content of them one frame behind the current.

Novum:
I went ahead and implemented what I suggested in a2ea879a1e7f4253da84e653ed56494667b64687, so I'm going to close this.

rg3:
I didn't know you were at id. Thanks for taking the time to review this. Your solution is really nice and not only fixes the problem of the new allocation not fitting in 1.5x the space, but also having to reallocate resources several times per frame. My best wishes for the incoming Doom Eternal launch. :+1:

PR #172, opened 2019-05-25
Support MacOS build with MoltenVK

Abuseware:


Abuseware:
Fullscreen mode goes insane, and window isn't "Retina" capable, but anything else works just fine.

PR #176, opened 2019-06-20
Fix push constant layout in R_ComputeWarpTexture

kvark:
I believe the intent here is to use the same pipeline layout as the one of the active pipeline.


Novum:
Indeed. I'm surprised the validation layers didn't catch this.

kvark:
O hey, so we did intersect before!

PR #179, opened 2019-06-23
add type 'float' to alpha to avoid implicit integer

rfht:
Noticed by me when compiling with clang 7 on OpenBSD, also got a report of gcc doing implicit int as well.


Novum:
Sure. This is harmless though, 1 is exactly representable as float and int.

PR #180, opened 2019-06-27
Wrong allocation size little fix proposal.

devnexen:


Novum:
All the handles have the same size, so this was harmless.

PR #206, opened 2020-02-11
Select UNORM format as surface format to avoid extra sRGB encoding

### LINKED ISSUE #161: Game is too bright, even with the lowest value
I can't lower contrast and brightness, the lowest values looks like this
image
image
This problem does not occour with Quakespasm, the colors are smooth: 
image
I'm using Radeon with Mesa drivers

davll:
Fixes #161 

Tested on
OS: Manjaro Linux (Arch-based) (kernel version = 4.19.101)
CPU: x86-64
GPU: Intel HD Graphics 620 (mesa)

### DIFF
--- a/Quake/gl_vidsdl.c
+++ b/Quake/gl_vidsdl.c
@@ -1543,14 +1543,27 @@ static qboolean GL_CreateSwapChain( void )
 		break;
 	}
 
+	// Select surface format
+	uint32_t surface_format_index = 0;
+	for (i = 0; i < format_count; ++i) {
+		if (surface_formats[i].format == VK_FORMAT_B8G8R8A8_UNORM ||
+		    surface_formats[i].format == VK_FORMAT_R8G8B8A8_UNORM ||
+			surface_formats[i].format == VK_FORMAT_B8G8R8_UNORM ||
+			surface_formats[i].format == VK_FORMAT_R8G8B8_UNORM )
+		{
+			surface_format_index = i;
+			break;
+		}
+	}
+
 	VkSwapchainCreateInfoKHR swapchain_create_info;
 	memset(&swapchain_create_info, 0, sizeof(swapchain_create_info));
 	swapchain_create_info.sType = VK_STRUCTURE_TYPE_SWAPCHAIN_CREATE_INFO_KHR;
 	swapchain_create_info.pNext = NULL;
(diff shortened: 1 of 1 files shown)


Novum:
No, the swap chain is supposed to be SRGB. I believe this is a bug in the Intel driver.

Novum:
So this is confusing to me because the color space is `VK_COLOR_SPACE_SRGB_NONLINEAR_KHR` on my Windows driver for both `VK_FORMAT_B8G8R8A8_UNORM` and `VK_FORMAT_B8G8R8A8_SRGB`. I didn't expect there to be another gamma correction for the SRGB format. Need to read a bit what the specified behavior is.

Novum:
I suspect that the `VK_FORMAT_B8G8R8A8_UNORM` format means there is an automatic conversion from linear to sRGB at the final step to the swap chain. Quake of course never did linear correct rendering, all intermediate values are already sRGB.

davll:
@Novum 

It is easily confusing that `VkSurfaceFormat` has two fields `format` and `colorSpace`. XD

According to [Chapter 29 of Vulkan 1.1 Specs](https://www.khronos.org/registry/vulkan/specs/1.1-khr-extensions/html/chap29.html):

> While the format of a presentable image refers to the encoding of each pixel, the `colorSpace` determines how the presentation engine interprets the pixel values. A color space in this document refers to a specific color space (defined by the chromaticities of its primaries and a white point in CIE Lab), and a transfer function that is applied before storing or transmitting color data in the given color space.

So, `colorSpace` tells how the driver "interprets" the **pixel values** of presentation images, and `format` denotes the way how the device encode **color values** into **pixel values**

The two suffixes `_SRGB` and `_UNORM` have different encoding and decoding operations for reading and writing pixels.

davll:
@Novum 

> I suspect that the VK_FORMAT_B8G8R8A8_UNORM format means there is an automatic conversion from linear to sRGB at the final step to the swap chain.

I don't think so, UNORM formats don't do any conversions besides `clamp(x * 0.5 + 0.5, 0.0, 1.0) * BITS`

Novum:
There is definitely a conversion going on, otherwise the result wouldn't be different.

Novum:
I fixed this more comprehensively in the meantime.

PR #216, opened 2020-04-18
Check only for VK_ERROR_FULL_SCREEN_EXCLUSIVE_MODE_LOST_EXT on Windows

laanwj:
VK_ERROR_FULL_SCREEN_EXCLUSIVE_MODE_LOST_EXT only exists on Windows so only check for it on Windows. Fixes the build on Linux.


Novum:
No. VK_ERROR_FULL_SCREEN_EXCLUSIVE_MODE_LOST_EXT is defined in the Vulkan headers independent of the platform. Your Vulkan headers are too old. Won't fix.

laanwj:
You're right, I was trying to compile on an ancient linux distro back then (that is, something derived from Debian stable :laughing:).

PR #223, opened 2020-05-20
update Arch dependencies

micwoj92:
This 'reverts #66'. These are needed to build package.


mikezila:
I can confirm that vkQuake 1.04.1 (which is the version up on the AUR currently) doesn't build on Arch without `opusfile` and `libmikmod` installed.

micwoj92:
@mikezila ugh i just realized than in makefile it is 
USE_CODEC_FLAC=0
USE_CODEC_OPUS=0
USE_CODEC_MIKMOD=0
and in aur these are set to 1

I just should not do things at night...

PR #226, opened 2020-07-03
Fix warp texture barrier dstAccess

jzulauf-lunarg:
The image layout transition barrier had the incorrect dstAccessMask,
guarding transfer *read* operations, for mip levels that are the
destination of the mip generation blits. Updated dstAccessMask to
transfer *write*.


jzulauf-lunarg:
@Novum -- found this with some new validation code I'm testing.

Novum:
Seems correct after short inspection

PR #231, opened 2020-08-22
Increase push constant size to avoid validation errors

karl-lunarg:
Per the Vulkan spec, every push constant value must be set prior to a
draw call.

Change the push constants calls in a couple of places to set all 21
float values in the basic_pipeline_layout.

### DIFF
--- a/Quake/gl_draw.c
+++ b/Quake/gl_draw.c
@@ -819,7 +819,7 @@ static void GL_OrthoMatrix(float left, float right, float bottom, float top, flo
 	float ty = (top + bottom) / (top - bottom);
 	float tz = -(f + n) / (f - n);
 
-	float matrix[16];
+	float matrix[21];
 	memset(&matrix, 0, sizeof(matrix));
 
 	// First column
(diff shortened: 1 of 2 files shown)


Novum:
Okay, thanks for finding this. I'm not quite happy with the fix though. I don't think GL_OrthoMatrix is the right place to do this

Novum:
I don't get any validation errors with SDK 1.2.148.1. Is that change not in the SDK yet?

karl-lunarg:
> I don't get any validation errors with SDK 1.2.148.1. Is that change not in the SDK yet?

It is not in the 148 SDK.  It would be in the next SDK, probably in the last half of September.

I was doing some work on the layers and was building it from the repo, using the latest code.

karl-lunarg:
> Okay, thanks for finding this. I'm not quite happy with the fix though. I don't think GL_OrthoMatrix is the right place to do this

Please feel free to discard the PR and fix it another way.

I tried to follow the logic through enough to have it work, but the fix does seem a bit fragile.  And I just started looking at the code enough to fix this and was not able to invest completely.

It occurred to me after making the PR that it may be clearer and more correct to assume that the push constant commands are pushing the right amount of data and instead change the layouts so that each pipeline has the correct push constant range.  That is, make the ranges match what is getting pushed instead of making all the ranges span the "worst case" of 21 floats.  This would result in pushing only the needed data instead of pushing all 21 with "dummy values" in place for data that isn't referenced.

I didn't look into the code deeply enough to understand how and when each layout is used, but perhaps you have a better understanding of this.

karl-lunarg:
I'm going to close this PR because the validation layer developers are still working on this part of the layer.  I've pointed them to this app to use as a test case.

The doubt I [mentioned](https://github.com/Novum/vkQuake/pull/231#issuecomment-678706528) here seems to make sense.  If you are not using a specific push constant in a shader, then you shouldn't have to write it.

A PR was added that was supposed to check to see if a push constant is used by a shader before complaining that it wasn't set.  But the layer is still reporting errors.

If it turns out that there is still an issue in vkQuake, we'll come back with a new issue or PR.

layer PRs:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/2106
https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/2138

PR #237, opened 2020-09-28
Add a package to list necessary to build on Ubuntu

Hurricos:
Just needed `libx11-xcb-dev`. Tested working now with a copy of Ubuntu Mate 20.04 -- sounds like Ubuntu updated their packages.


Hurricos:
@Novum This one.

Again, it builds and runs without issue now on Ubuntu 20.04. This just adds one dep to the list of packages to build on a desktop installation of Ubuntu 20.04.

PR #248, opened 2020-10-15
Add basic bug report template

micwoj92:
Not perfect, but should help in case something was already fixed on master branch but not in release or if something is OS specific bug.


Novum:
I really don't mind freeform bug reports. Templates don't make people magically produce more useful descriptions.

PR #253, opened 2020-10-28
added path for AUR build

mardab:
Tested current AUR build, putting pak0 into any of vkQuake folders in /usr/ does not work, so the next best thing is to have id1 folder in home directory.

Additionally, Ubuntu path had slashes the wrong way, so I've corrected it.

For some unknown to me reason, VSC insists on these changes to it's config file, I'm pretty sure these only apply to my case, but don't know how to make VSC not do that.

### DIFF
--- a/.vscode/c_cpp_properties.json
+++ b/.vscode/c_cpp_properties.json
@@ -21,7 +21,10 @@
             "macFrameworkPath": [
                 "/System/Library/Frameworks",
                 "/Library/Frameworks"
-            ]
+            ],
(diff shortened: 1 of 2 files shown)


micwoj92:
Hello, I maintain AUR package and when you install it should say in terminal that you have to put it there. If you compile compile using steps from readme then the ubuntu path is good.

https://aur.archlinux.org/cgit/aur.git/tree/vkquake.install?h=vkquake

mardab:
I see. Anyway, that should be put in README, just in case.

PR #266, opened 2020-12-06
Made it compile under macOS 10.15 (Catalina)

danieleftodi:
Hi,

Made it compile under macOS 10.15 (Catalina)

cheers,
daniel.


Quake/Makefile
```diff
@@ -176,7 +176,7 @@ endif
 COMMON_LIBS:= -lm
 
 ifneq ($(HOST_OS),darwin)
-COMMON_LIBS+= -lX11-xcb -lvulkan
+COMMON_LIBS+= -lX11-xcb -lvulkan -lMoltenVK
```
sezero, line 179: Why are you adding -lMoltenVK to the non-darwin case and breaking linux, etc builds?

Quake/gl_vidsdl.c
```diff
@@ -417,10 +417,19 @@ static qboolean VID_SetMode (int width, int height, int refreshrate, int bpp, qb
 		if (vid_borderless.value)
 			flags |= SDL_WINDOW_BORDERLESS;
 		
-		draw_context = SDL_CreateWindow (caption, SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED, width, height, flags);
+		draw_context = SDL_CreateWindow (
```
sezero, line 418: Why is this style difference necessary?

Shaders/compile.sh
```diff
@@ -1,26 +1,26 @@
 #!/bin/bash
 
 #set your VULKAN_SDK location  before running
-VULKAN_SDK=~/VulkanSDK/1.0.26.0/x86_64/
+#VULKAN_SDK=~/VulkanSDK/1.0.26.0/x86_64/
```
sezero, line 4: Instead of commenting this out, you can conditionalize it by adding an environment check around it

.gitignore
```diff
@@ -13,3 +13,44 @@ Shaders/Compiled/*spv
 Shaders/bintoc.exe
 Shaders/bintoc.obj
 Windows/VisualStudio/.vs
+
+.DS_Store
+
+
```
sezero, line 19: The ignoring and removal of these generated files is very rude. One would prefer them to be present in the source tree.

Quake/Makefile
danieleftodi: @sezero yeah, *opps* sorry ... it was actually a long time ago I did this. So now that I look back on it, yeah, why did I do that? 🤔

Fixing it in the next commit.

Quake/gl_vidsdl.c
danieleftodi: a little bit force of habit I'm following since making Doom 3 (dhewm3), vkQuake3 and RBDOOM-3-BFG (that I'm currently working on fixing to run on macOS...) ... as in them, that is the style guide. 

And when I traced/read the code, It just happened to catch my eye and tweaked it :-)

Also, IMHO, usually it's easier to following this style guide in general, as sometimes, some names can become quite long ... so being able to read one per line, usually makes it fit in our window.

Shaders/compile.sh
danieleftodi: Sure, we could do that. But this should be instead done / handled by default by the Vulkan SDK (and the user should follow their instructions into how to install it and setup their Vulkan SDK)

As per the official Vulkan SDK, they provide a very neat and clean way for this. They recommend adding the following to your "~/.bash_profile"

`# Vulkan SDK Stuff
source /opt/vulkan-sdk/USING/setup-env.sh
`

.gitignore
danieleftodi: I have made a fix for this, in the next commit. Since shaders are platform independent (and gpu interchangeable) it make sense to provide pre-compiled shaders to endusers.

Though, I'm leaving the binary "bintoc" in the .gitignore ... since it's not platform interchangeable .. and the compile.sh scripts handles it, if it's missing by compiling one for the users platform. 

Also, it would be best if each user from time to time re-compiled them on their own ... as the Vulkan SDK advances and better and maybe more efficient compiled shaders can be produces. I will add this into a build script, to re-compile them, and then the source code.

Quake/gl_vidsdl.c
Novum: Please revert

.gitignore
Novum: You have a misconception there. I don't use glslang to optimize. The SPIR-V is a pretty literal translation from GLSL. Optimization should be done by the driver. The glslang optimization options just exist because of absurdly garbage mobile drivers and are counter productive on PC.

sezero: The generated `*.c` are still removed in the latest version of your patch.

danieleftodi: > The generated `*.c` are still removed in the latest version of your patch.

They should be, please see: https://github.com/Novum/vkQuake/pull/266/commits/149c6df7f56eab2ecfa9182c47c0aa77140da2e1

danieleftodi: > You have a misconception there. I don't use glslang to optimize. The SPIR-V is a pretty literal translation from GLSL. Optimization should be done by the driver. The glslang optimization options just exist because of absurdly garbage mobile drivers and are counter productive on PC.

👍🏻 ... thanks for clarifying it 🙂 *happy-to-learn*

Though, I'm wondering a bit how Apple will treat shaders with their new M1, and future to come gpu's ... since they do have a "if you can, please optimise your code to make it fly on our hardware" ... I don't know, but felt it's better to make a build script that re-compiles them before each build (in case, in the future, things change... and re-compile is beneficiary...)

Quake/gl_vidsdl.c
danieleftodi: > Please revert

done 👍🏻 (https://github.com/Novum/vkQuake/pull/266/commits/d4590bc4fd22514a1e633a688ccd0c855a857c79)

sezero:
OK, I did a `wget https://github.com/Novum/vkQuake/pull/266.patch`
(which is about 475k), applied it locally, and did a `git diff -w`
which resulted in the attached 14k patch: [patch.txt](https://github.com/Novum/vkQuake/files/5654062/patch.txt)

Basically it only updates the README and does a minor tweak to the
shader compile.sh script. _However_, the compiled shaders do differ:
is that intended?

.gitignore
Novum: It's not beneficiary. Drop it.

sezero: Honsestly, I don't know how this PR is beneficial in whole: https://github.com/Novum/vkQuake/pull/266#issuecomment-740045611

micwoj92:
> OK, I did a `wget https://github.com/Novum/vkQuake/pull/266.patch`
> (which is about 475k), applied it locally, and did a `git diff -w`
> which resulted in the attached 14k patch: [patch.txt](https://github.com/Novum/vkQuake/files/5654062/patch.txt)
> 
> Basically it only updates the README and does a minor tweak to the
> shader compile.sh script. _However_, the compiled shaders do differ:
> is that intended?

look how weirdly GitHud did diff for the shaders

Novum:
Going to close this. Not happy with the patch.

PR #283, opened 2021-01-24
Load external .vis files

### LINKED ISSUE #242: [Feature request] .vis file support
It would be great if .vis files were supported for transparent water. Like this, the lit/vis packs from Mark V could be used directly and full water transparency would be possible without having to use r_novis 1 or vis-patched maps.

daniel-abbott:
Resolves #242 

Loads .vis files from /maps, functionally identical to Mark V Quake.

Also contains two changes to .gitignore to avoid committing id1 from where I'm using it, and my console history.

### DIFF
--- a/Quake/gl_model.c
+++ b/Quake/gl_model.c
@@ -1989,54 +1989,11 @@ typedef struct vispatch_s
 // 2001-12-28 .VIS support by Maddes  end
 #define VISPATCH_HEADER_LEN_36 36
 
-///*
-//=================
-//Mod_LoadEntities
-//=================
-//*/
-//void Mod_LoadEntities(lump_t* l)
-//{
-//	char	entfilename[MAX_QPATH];
-//	char* ents;
-//	int		mark;
-//	unsigned int	path_id;
-//
-//	if (!external_ents.value)
-//		goto _load_embedded;
-//
-//	q_strlcpy(entfilename, loadmodel->name, sizeof(entfilename));
-//	COM_StripExtension(entfilename, entfilename, sizeof(entfilename));
-//	q_strlcat(entfilename, ".ent", sizeof(entfilename));
-//	Con_DPrintf2("trying to load %s\n", entfilename);
-//	mark = Hunk_LowMark();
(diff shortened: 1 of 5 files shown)


Quake/gl_model.c
```diff
@@ -1917,6 +1946,142 @@ void Mod_LoadSubmodels (lump_t *l)
...
+
+typedef struct vispatch_s
+{
+	char	mapname[VISPATCH_MAPNAME_LENGTH];	// Baker: DO NOT CHANGE THIS to MAX_QPATH_64, must be 32
+	int		filelen;		// length of data after VisPatch header (VIS+Leafs)
+} vispatch_t;
+// 2001-12-28 .VIS support by Maddes  end
+#define VISPATCH_HEADER_LEN_36 36
+
+///*
```
Novum: Why the commented out code?

Novum: Nevermind, was removed later.

Novum:
Can we maybe drop the author name comments? I'm personally not a big fan of that.

Novum:
I had to revert this because it broke arcane dimensions.

Novum:
Not quite sure why though. It was complaining about invalid leafs. I need to investigate but wanted to release 1.05.3 as soon as possible because of the dynamic lights fix.

sezero:
In case there still is interest in this, the attached patch fixes a lot
of bugs found in the original and simplifies it: [283_new_patch.txt](https://github.com/Novum/vkQuake/files/6835981/283_new_patch.txt)

- Fixes `path_id` handling, which seems to be the root cause of
  Arcane Dimensions issues. (the original not only didn't respect
  the path_id, but also clobbered the path_id of the map itself)
- Doesn't allow external vis files for bsp2 maps.
- Frees hunk if vis file loading fails in the middle.
- Fixes wrong FS_flcose use.
- Simplifies the whole patch.

sezero:
Pushed https://github.com/Novum/vkQuake/commit/b891ecbd1d66146ad611cb2a4d5d159d67243c36 as a follow-up patch

PR #304, opened 2021-06-16
minor warning fixes:

sezero:
- fix comparison between integer and pointer in gl_texmgr.c in 32 bits:
  VkImageView is an uint64_t, not a pointer in 32 bits.
- make sure alloca() is defined in glquake.h for R_BindPipeline().


Novum:
Thank you.
be01e225cbbc5cfd1b3a61b3894e5ec8586a2ca2
Slightly different, I prefer explicit comparisons.

Issue #316, opened 2021-06-27
Broken huds in the majority of map packs


El-Juancho:
**RAVENKEEP**
![image](https://user-images.githubusercontent.com/70743149/123552815-3b8bd800-d778-11eb-93f2-792a0274b929.png)
**XMEN**
![image](https://user-images.githubusercontent.com/70743149/123552832-565e4c80-d778-11eb-899e-36c598993e48.png)
**Q25LIMIT**
![image](https://user-images.githubusercontent.com/70743149/123552852-7d1c8300-d778-11eb-9ac4-4744d7e96087.png)
**DWELL**
![image](https://user-images.githubusercontent.com/70743149/123552875-98878e00-d778-11eb-808c-9716d5c6df36.png)

The end screen is broken too
![image](https://user-images.githubusercontent.com/70743149/123552960-0df35e80-d779-11eb-8b57-9af6ae350362.png)

Novum:
Duplicate  #313

JuanchoES:
> Duplicate #313

I did not know that is the same bug, sorry

Novum:
I cannot reproduce this on either AMD nor NVIDIA drivers on Windows. What configuration is this?

JuanchoES:
Oops, i tested the new cvar "cl_nocsqc 1" in arcane dimensions, it does work but breaks the hud when playing other map packs, my problem is fixed when i write "cl_nocsqc 0" in the console

Novum:
Okay, that's still kind of a bug then. Reopening

JuanchoES:
I hope this helps

https://drive.google.com/file/d/1KxhVGaLQ63xfN3Uuj32b4y1BE7r_uqJR/view?usp=sharing

JuanchoES:
When using these launch parameters my hud is still broken, this is really weird

![image](https://user-images.githubusercontent.com/70743149/123560798-cedb0280-d7a4-11eb-9569-0680439b86a8.png)
![image](https://user-images.githubusercontent.com/70743149/123560801-d7cbd400-d7a4-11eb-983c-b934c94bee5b.png)

Novum:
First bad commit is 4993ee35de5f139e3b44b1db3edbdb5d3be20c6d

Novum:
42522db3b255bcc921b3d8b55a858bd3a3687b3d

andrei-drexler:
It looks like this one isn't quite dead yet - starting vkQuake with `-game ad` and typing in the console
> map start
[...]

PR #318, opened 2021-06-28
Fix broken HUDs on game change

andrei-drexler:
Fix for issue #316 (https://github.com/Shpoike/Quakespasm/commit/683d0e67c70857eb19d20f00f96d0f3085f9bfaa)

### DIFF
--- a/Quake/gl_draw.c
+++ b/Quake/gl_draw.c
@@ -430,16 +430,16 @@ void Draw_NewGame (void)
 
 	Scrap_Upload (); //creates 2 empty gltextures
 
+	// empty lmp cache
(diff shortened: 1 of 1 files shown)


Novum:
You have commit rights btw. If it's simple fixes like this I'm fine with just pushing them.

andrei-drexler:
Alright, will do! I wasn't sure where the line is, so I only felt comfortable pushing SIMD-related changes. Otherwise, even for simple fixes I thought it's best to assume that I might not know the full story and open PR's that you can review instead of pushing.

Novum:
No line, this isn't mission critical infrastructure or something. If something is not ideal it can be reverted or changed.

PR #329, opened 2021-07-25
Fix startup crash ("failed to find vkSetDebugUtilsObjectNameEXT") in debug builds

Macil:
When I tried to run a debug build on MacOS, it would crash on startup with the error "QUAKE ERROR: vkGetDeviceProcAddr failed to find vkSetDebugUtilsObjectNameEXT". I [found a post with an explanation and a fix for the issue](https://forums.developer.nvidia.com/t/vksetdebugutilsobjectnameext-causes-a-crash-when-attempting-to-set-a-name-for-vkphysicaldevice/79517/6) so I followed its instructions and the issue appeared fixed. It seems like the previous code only worked on some platforms because of [some workarounds in specific platforms](https://github.com/KhronosGroup/Vulkan-Loader/issues/116#issuecomment-539176985). It seems like this change makes it work fully following the standards now.

### DIFF
--- a/Quake/gl_vidsdl.c
+++ b/Quake/gl_vidsdl.c
@@ -176,6 +176,11 @@ static uint32_t current_swapchain_buffer;
 	if (fp##entrypoint == NULL) Sys_Error("vkGetInstanceProcAddr failed to find vk" #entrypoint); \
 }
 
+#define GET_GLOBAL_INSTANCE_PROC_ADDR(_var, entrypoint) { \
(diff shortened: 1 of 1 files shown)


Novum:
Does this run on Linux/Windows like this?

Novum:
Btw. I think Mac support is probably a good idea because QuakeSpasm uses OpenGL which will likely disappear at some point on Macs.

sezero:
This made me go past the extension load failure under Linux, that much is good,
but I got stuck with a segfault later - see https://github.com/Novum/vkQuake/issues/331

Macil:
I tested this on Windows and it worked there too.

PR #330, opened 2021-07-25
Build config improvements for MacOS

Macil:
- The MacOS instructions in the readme don't work as-is on newer M1 ARM Macs because homebrew now installs into /opt/homebrew instead of /usr/local and clang doesn't know to look there by default. Since installing the libraries through homebrew is the readme-recommended setup and is what most Mac devs are probably doing, I think it's worth supporting out of the box like this. It won't have any downsides to people not using homebrew, or people using homebrew but not for these libraries.
- The VSCode launch config was changed to start and debug vkquake with the CodeLLDB extension [because the VSCode default doesn't support debugging on M1 Macs currently](https://github.com/microsoft/vscode-cpptools/issues/6779#issuecomment-854675319). I assume that will work well for other platforms too but if people had any attachment to the old setup then this launch task could be moved into a new "vkquake (Mac)" launch task instead.

### DIFF
--- a/Quake/Makefile
+++ b/Quake/Makefile
@@ -152,6 +152,10 @@ ifneq ($(HOST_OS),darwin)
 COMMON_LIBS+= -lvulkan
 CFLAGS+= -DLINUX
 else
+ifeq ($(shell test -d /opt/homebrew && echo true),true)
+COMMON_LIBS+= -L/opt/homebrew/lib
+CFLAGS+= -I/opt/homebrew/include
+endif
 COMMON_LIBS+= -lMoltenVK
 endif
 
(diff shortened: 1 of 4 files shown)


Quake/Makefile
```diff
@@ -153,6 +153,10 @@ COMMON_LIBS+= -lvulkan
 CFLAGS+= -DLINUX
 else
 COMMON_LIBS+= -lMoltenVK
+ifeq ($(shell test -d /opt/homebrew && echo true),true)
```
sezero, line 156: Maybe move these just before `-lMoltenVK` ?

Macil: I reordered them now.

Novum:
I have a hard time judging this because I don't use Macs. @sezero do you have an idea?

sezero:
Only by eyeballing, it looks good to me (I don't have a vulkan-capable mac.)

Issue #332, opened 2021-08-01
Differences in sound, and distant dithering in Ter_Shibboleth_Drake when compared to FTEQW


ZecChelovek:
Dithering, first screenshot is vkQuake:
https://imgur.com/a/04LFd4A

I don't have time to record the sound right now, but when shib1_drake is fired up in the latest vkQuake, the ambient noise seems to be running twice over itself. This issue isn't present in FTEQW.

Novum:
Please check against QuakeSpasm/QSS, not FTEQW.

ZecChelovek:
Top is QS, bottom is QSS. No distant dithering.
https://imgur.com/a/VzQvmEu

Still no time to record sound, but the double sound is present in QS and QSS. I suppose it's something to do with the map itself.

Novum:
Where can I download this map?

Novum:
Okay got it to run. This is because the skybox geo and the map geo are coplanar and there is z-fighting.

Won't fix. This would only be fixable by reverting a pretty substantial optimization for skybox rendering on the GPU. Not worth it for one map.

andrei-drexler:
@Novum any thoughts on switching to a reversed-z projection matrix? I gave it a quick try and it seemed to eliminate the distant z-fighting on that map even with the default D24_S8 buffer. We could even switch to an infinite far plane and get rid of `gl_farclip` altogether (TerShibboleth's quake.rc sets it to 512K!), and if we made D32F_S8 the preferred format we probably wouldn't have to worry about this ever again <sup>1</sup>

<sup>1</sup> famous last words

Novum:
As far as I know reverse depth shouldn't really help if not using a float z buffer?

andrei-drexler:
I haven't done any rigorous precision tests myself, but according to [this blog post](https://dev.theomader.com/depth-precision/#:~:text=Reverse%20depth%20seems%20to%20perform%20slightly%20better%20than%20standard%20depth%2C%20especially%20towards%20the%20far%20plane.)
> with integer buffers [...] reverse depth seems to perform slightly better than standard depth, especially towards the far plane

and in this case *slightly better* is apparently enough to eliminate the problem.

PR #336, opened 2021-08-06
Switch to reverse depth projection

andrei-drexler:
Fixes issue #332

### DIFF
--- a/Quake/gl_rmain.c
+++ b/Quake/gl_rmain.c
@@ -270,11 +270,11 @@ static void GL_FrustumMatrix(float matrix[16], float fovx, float fovy)
 	matrix[1*4 + 1] = -h;
 	
 	// Third column
-	matrix[2*4 + 2] = -f / (f - n);
+	matrix[2*4 + 2] = f / (f - n) - 1.0f;
 	matrix[2*4 + 3] = -1.0f;
 
 	// Fourth column
-	matrix[3*4 + 2] = -(n * f) / (f - n);
+	matrix[3*4 + 2] = (n * f) / (f - n);
 }
 
 /*
@@ -484,7 +484,7 @@ void R_DrawViewModel (void)
 				gly + glheight - r_refdef.vrect.y - r_refdef.vrect.height,
 				r_refdef.vrect.width,
 				r_refdef.vrect.height,
-				0.0f, 0.3f);
+				0.7f, 1.0f);
 	
 	R_DrawAliasModel (currententity);
 
--- a/Quake/gl_rmisc.c
+++ b/Quake/gl_rmisc.c
@@ -1581,7 +1581,7 @@ void R_CreatePipelines()
 	depth_stencil_state_create_info.sType = VK_STRUCTURE_TYPE_PIPELINE_DEPTH_STENCIL_STATE_CREATE_INFO;
 	depth_stencil_state_create_info.depthTestEnable = VK_FALSE;
 	depth_stencil_state_create_info.depthWriteEnable = VK_FALSE;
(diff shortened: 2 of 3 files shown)


Novum:
Still not convinced. Without a float depth buffer this just shifts around precision issues?

Novum:
I think if we do reverse depth we should check for float depth support as a condition

andrei-drexler:
Intuitively, I expected it to be a wash, so I was surprised to see just how much of a difference this makes in practice even with an int buffer. If you load up the map from #332, the distance at which z-fighting becomes apparent is much, much higher with reversed depth, so much so that you'll need to use host_timescale 10 when noclipping, otherwise it will take a long time to reach that point. Of course, that's one map with one particular depth distribution. For a more general explanation, I think one key insight (and something that didn't occur to me at all until I read [this blog post](https://www.reedbeta.com/blog/depth-precision-visualized/) by Nathan Reed) is that distinct depths getting mapped to the same final value due to limited precision is *not* the only type of error encountered with the standard depth projection. There are also monotonicity errors: two depths can get mapped to different depth buffer values, but with the order reversed. These errors make z-fighting much more evident when moving the camera, and instability was mentioned as a major difference between the two setups in this [SIGGRAPH 2012 presentation](http://www.humus.name/Articles/Persson_CreatingVastGameWorlds.pdf#page=14&zoom=page-fit,-512,720) from Avalanche:

> When comparing z-fighting prone cases with regular and reversed fixed point buffers, in practice the reverse has consistently resulted in better quality. It has a slight edge in the amount of saw-tooth artifacts, but the primary difference is the stability. In problematic areas the reversed fixed point buffer will still exhibit some artifacts, but they will be stable and thus less objectionable. A regular depth buffer will flicker in motion between about the same artifacts as the reversed to something far worse. A stable saw-tooth pattern in the distance may go unnoticed by a player unless it is pointed out, but a flickering mess will instantly catch 
the attention and is hard not to notice. 

The same presentation claims that the standard projection also exhibits floating-point errors in the depth calculations even before quantization to the fixed-point depth buffer:
> It is a somewhat unintuitive result that reversing the depth buffer also helps with a fixed point depth buffer considering the resolution is uniform and there is a unique 32bit float matching every value in 24bit fixed point. The reason this is the case is that plenty of the error that can be introduced in the process happens before the rasterization, in other words, you get poor inputs to the rasterizer from the vertex shader and the lowest bits would for all practical purposes be noise. As we saw a few slides back, we are really operating on the edge of the precision of floats here when you need 24bit output, and any sort of math you do will likely lose a bit or two before you are done, even if we are careful about it. With a reversed projection matrix the math will mostly operate in a range that is much more comfortable for floats, i.e. close to zero, and the rasterizer is given more precise values as inputs. This is also why flipping the depth range should be done with the projection matrix, not with the viewport, because at that point it is already too late.

> We actually shipped with 24bit fixed point buffer on PC, because DX10 does not have a D24F format that the consoles have, and we needed stencil. We could have gone with D32F_S8_X24, but since 24bit was enough we stuck with that.

Nathan Reed's conclusion:
> In any perspective projection situation, just use a floating-point depth buffer with reversed-Z! And if you can’t use a floating-point depth buffer, you should still use reversed-Z.

If both formats are supported we could make D32F the preferred one, but IMO we should probably test this on a bandwidth-starved APU/laptop GPU. On an old R5 M255, switching the color attachments from LOAD_OP_CLEAR to LOAD_OP_DONT_CARE increased the framerate by up to 5%, and I'm guessing using a fatter z+stencil buffer would similarly not go unnoticed.

Novum:
I'll reply in depth, but one thing is that Radeons only support F32 depth internally, F24 is actually emulated and can be slower if you read values because of conversion.

NVIDIA actually benefits very slightly from 24 bit depth, but it's not a massive thing either.

Issue #319, opened 2021-06-30
Wrong liquid texture coordinates


c-d-a:
vkQuake's warptextures are flipped vertically compared to Quakespasm.
But Quakespasm's own coordinates inherit the incorrect behavior from GLQuake (shifted by 1/2 on both axes, as compared to DOS Quake & WinQuake).

WinQuake:
![winq](https://user-images.githubusercontent.com/55441216/123977423-6aa08480-d9d8-11eb-99a8-f72e0655408f.png)
GLQuake:
![glq](https://user-images.githubusercontent.com/55441216/123977451-6d9b7500-d9d8-11eb-928a-8350dcad7108.png)
Quakespasm:
![qs](https://user-images.githubusercontent.com/55441216/123977459-6ffdcf00-d9d8-11eb-9283-479853e32ea5.png)
vkQuake:
![vk](https://user-images.githubusercontent.com/55441216/123977470-71c79280-d9d8-11eb-8f00-fa54afee3a23.png)

[Test map](https://github.com/Novum/vkQuake/files/6741622/warp_co.zip).

LethalManBoob:
This is very interesting. I hope this was sent to quakespasm as well

Novum:
Is there actually a practical map where this is visible?

c-d-a:
Well, I discovered this while trying to do a Doom-style wall of faces.
But I can't name any existing maps using discernible patterns as liquids, no.

Novum:
I'm usually trying to match QuakeSpasm

c-d-a:
Well, following QS would at least make the direction consistent, better than nothing.

Anyway, I have fiddled with this some more. The issue goes deeper than a simple half-shift. WinQuake can project the texture more or less seamlessly across faces with roughly the same direction. Meanwhile, GLQ/QS seem to come up with a different set of disjointed coordinates for each face. But TyrGLQuake, for example, seems to have this solved.
So this affects anyone who had ever tried making a fancy exit portal, or a circular force field, or a glass tube with water, or a floating lava ball.

https://user-images.githubusercontent.com/55441216/124144737-a9e8d700-daa5-11eb-86ec-5aa1bcdd8791.mp4

https://user-images.githubusercontent.com/55441216/124144785-b40ad580-daa5-11eb-8b0b-beab4e947c82.mp4

https://user-images.githubusercontent.com/55441216/124144813-ba00b680-daa5-11eb-8876-2a2e3ffc0d52.mp4

[...]

PR #339, opened 2021-08-08
Reset pointers to sky textures on map unload/game change

andrei-drexler:
After loading a map with a sky (e.g. start), switching to a skyless one (e.g. warp_co from #319) leads to a crash in `Sky_DrawSkyLayers` due to stale pointers to the solid/alpha sky textures (freed in `TexMgr_FreeTexturesForOwner`, called by `Mod_ClearAll`).

The same scenario doesn't crash QS because QS doesn't try to use the textures until after the [bounds check](https://sourceforge.net/p/quakespasm/quakespasm/ci/3fc32bf7fe93d40d63bd0f48df67467d7f8efd25/tree/Quake/gl_sky.c#l966) (which in this case always fails), whereas vkQuake first binds the pipeline and the (stale) texture descriptors and then proceeds to the bound check and the actual drawing:
https://github.com/Novum/vkQuake/blob/5bf976a4afc468b1873316725132e1c7597f15eb/Quake/gl_sky.c#L909-L916
A minimal solution would have been to delay the bindings until they are first needed, or maybe move them to `Sky_DrawFace`, but clearing the pointers on map unload/game change seems more robust that relying on just the bounds check.

@sezero can you take a look at this, please? I'm hoping it could also be integrated into QS (even if not strictly necessary) to avoid further divergence.

### DIFF
--- a/Quake/common.c
+++ b/Quake/common.c
@@ -2563,6 +2563,7 @@ static void COM_Game_f (void)
 		//clear out and reload appropriate data
 		Cache_Flush ();
 		Mod_ResetAll();
+		Sky_ClearAll();
 		if (!isDedicated)
 		{
 			TexMgr_NewGame ();
--- a/Quake/gl_model.c
+++ b/Quake/gl_model.c
@@ -228,7 +228,6 @@ void Mod_ClearAll (void)
 			mod->needload = true;
 			TexMgr_FreeTexturesForOwner (mod); //johnfitz
 		}
-	Sky_ClearAll();
 }
 
 void Mod_ResetAll (void)
@@ -246,7 +245,6 @@ void Mod_ResetAll (void)
 		memset(mod, 0, sizeof(qmodel_t));
 	}
 	mod_numknown = 0;
-	Sky_ClearAll();
 }
 
 /*
--- a/Quake/gl_model.c
+++ b/Quake/gl_model.c
@@ -228,6 +228,7 @@ void Mod_ClearAll (void)
 			mod->needload = true;
 			TexMgr_FreeTexturesForOwner (mod); //johnfitz
 		}
+	Sky_ClearAll();
 }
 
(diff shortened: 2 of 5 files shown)


sezero:
This should be good, tried it in qs with no ill effects.
However, don't hide Sky_ClearAll calls in Mod_ResetAll or Mod_ClearAll:
I will do it in qs like the following:
```diff
diff --git a/Quake/common.c b/Quake/common.c
index a449001..68005f9 100644
--- a/Quake/common.c
+++ b/Quake/common.c
@@ -2201,6 +2201,7 @@ static void COM_Game_f (void)
 		//clear out and reload appropriate data
 		Cache_Flush ();
 		Mod_ResetAll();
+		Sky_ClearAll();
 		if (!isDedicated)
 		{
 			TexMgr_NewGame ();
diff --git a/Quake/host.c b/Quake/host.c
index ad60160..9261e72 100644
--- a/Quake/host.c
+++ b/Quake/host.c
@@ -541,6 +541,7 @@ void Host_ClearMemory (void)
 	Con_DPrintf ("Clearing memory\n");
 	D_FlushCaches ();
 	Mod_ClearAll ();
+	Sky_ClearAll ();
 /* host_hunklevel MUST be set at this point */
 	Hunk_FreeToLowMark (host_hunklevel);
 	cls.signon = 0;
```

andrei-drexler:
Alright, thanks! The reason those calls were in Mod_ClearAll/Mod_ResetAll is that I wanted the cause (TexMgr_FreeTexturesForOwner unloading the textures) and the effect (Sky_ClearAll resetting the pointers) to be as close as possible. Plus, Sky_LoadTexture is called from Mod_LoadTextures, so I thought clearing the pointers from Mod_* functions would also be somewhat symmetric. Either way, I've moved the calls, as suggested.

PR #344, opened 2021-08-22
Nightdive support

Perpixel:
Workaround for crashes when using the RERELEASE paks. Still need to work on getting what appears to be localized strings to fetch the right data.

This is just for reference.

### DIFF
--- a/Quake/Makefile
+++ b/Quake/Makefile
@@ -293,6 +293,7 @@ OBJS := strlcat.o \
 	sv_user.o \
 	world.o \
 	zone.o \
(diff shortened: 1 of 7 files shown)


sezero:
52 is `svc_achievement`

11 is `MOVETYPE_GIB`: maybe map to `MOVETYPE_BOUNCE`?

LethalManBoob:
Great to see this. The coloured lightmaps are somthing I cant come back from. :)

sezero:
Handled differently in current git. Also see PR #345.

PR #345, opened 2021-08-22
Add support for the 2021 re-release

andrei-drexler:
This fixes crashes with the 2021 re-release and adds basic support for localized strings so that the new maps display proper messages instead of text id's. Localization data is read from a `loc_english.txt` file in the mod directory, in the same format as the files in the `localization` folder in the `rerelease/QuakeEX.kpf` zip archive. Currently the files need to be copied manually - adding code to read zip files just for this feature seemed like feature creep, and can be added later if necessary.

@sezero, I'm hoping this can also be merged into QS, can you please have a look?

### DIFF
--- a/Quake/cl_parse.c
+++ b/Quake/cl_parse.c
@@ -1923,7 +1923,7 @@ void CL_ParseServerMessage (void)
 			cl.completed_time = cl.time;
 			vid.recalc_refdef = true;	// go to full screen
 			//johnfitz -- log centerprints to console
-			str = LOC_GetString (MSG_ReadString ());
+			str = MSG_ReadString ();
 			SCR_CenterPrint (str);
 			Con_LogCenterPrint (str);
 			//johnfitz
@@ -1934,7 +1934,7 @@ void CL_ParseServerMessage (void)
 			cl.completed_time = cl.time;
 			vid.recalc_refdef = true;	// go to full screen
 			//johnfitz -- log centerprints to console
-			str = LOC_GetString (MSG_ReadString ());
+			str = MSG_ReadString ();
 			SCR_CenterPrint (str);
 			Con_LogCenterPrint (str);
 			//johnfitz
--- a/Quake/cl_parse.c
+++ b/Quake/cl_parse.c
@@ -1923,7 +1923,7 @@ void CL_ParseServerMessage (void)
 			cl.completed_time = cl.time;
 			vid.recalc_refdef = true;	// go to full screen
 			//johnfitz -- log centerprints to console
-			str = MSG_ReadString ();
+			str = LOC_GetString (MSG_ReadString ());
 			SCR_CenterPrint (str);
 			Con_LogCenterPrint (str);
 			//johnfitz
@@ -1934,7 +1934,7 @@ void CL_ParseServerMessage (void)
 			cl.completed_time = cl.time;
 			vid.recalc_refdef = true;	// go to full screen
 			//johnfitz -- log centerprints to console
-			str = MSG_ReadString ();
+			str = LOC_GetString (MSG_ReadString ());
 			SCR_CenterPrint (str);
 			Con_LogCenterPrint (str);
 			//johnfitz
--- a/Quake/common.c
+++ b/Quake/common.c
@@ -2923,7 +2923,7 @@ fail:		if (fp) fclose(fp);
 	cursor = localization.text;
 
 	// skip BOM
-	if (cursor[0] == 0xEF && cursor[1] == 0xBB && cursor[2] == 0xB)
+	if ((unsigned char)(cursor[0]) == 0xEF && (unsigned char)(cursor[1]) == 0xBB && cursor[2] == 0xB)
 		cursor += 3;
 
 	lineno = 0;
@@ -2957,6 +2957,7 @@ fail:		if (fp) fclose(fp);
 			char *key_end = equals;
 			qboolean leading_quote;
 			qboolean trailing_quote;
+			locentry_t *entry;
 			char *value_src;
 			char *value_dst;
 			char *value;
@@ -3033,7 +3034,7 @@ fail:		if (fp) fclose(fp);
 				localization.entries = (locentry_t*) realloc(localization.entries, sizeof(*localization.entries) * localization.maxnumentries);
 			}
 
-			locentry_t *entry = &localization.entries[localization.numentries++];
+			entry = &localization.entries[localization.numentries++];
 			entry->key = line;
 			entry->value = value;
 		}
--- a/Quake/common.c
+++ b/Quake/common.c
@@ -2572,6 +2572,7 @@ static void COM_Game_f (void)
 		}
 		ExtraMaps_NewGame ();
 		DemoList_Rebuild ();
+		LOC_Init ();
 
 		Con_Printf("\"game\" changed to \"%s\"\n", COM_GetGameNames(true));
 
@@ -2839,3 +2840,332 @@ long FS_filelength (fshandle_t *fh)
 	return fh->length;
 }
 
+/*
+============================================================================
+								LOCALIZATION
+============================================================================
+*/
+typedef struct
+{
+	char *key;
+	char *value;
+} locentry_t;
+
+typedef struct
+{
+	int			numentries;
+	int			maxnumentries;
+	int			numindices;
+	unsigned	*indices;
+	locentry_t	*entries;
+	char		*text;
+} localization_t;
+
+localization_t localization;
+
+/*
+================
+COM_HashString
+Computes the FNV-1a hash of string str
+================
+*/
+unsigned COM_HashString (const char *str)
+{
+	unsigned hash = 0x811c9dc5u;
+	while (*str)
+	{
+		hash ^= *str++;
+		hash *= 0x01000193u;
+	}
+	return hash;
+}
+
+/*
+================
+LOC_LoadFile
+================
+*/
+void LOC_LoadFile (const char *path)
+{
+	// clear existing data
+	if (localization.text)
+	{
+		free(localization.text);
+		localization.text = NULL;
+	}
+	localization.numentries = 0;
+	localization.numindices = 0;
+
+	if (!path)
+		return;
+
+	localization.text = (char*) COM_LoadMallocFile(path, NULL);
+	if (!localization.text)
+	{
+		Con_DPrintf("Couldn't load localization file '%s'\n", path);
+		return;
+	}
+
+	char *cursor = localization.text;
+
+	// skip BOM
+	if (cursor[0] == 0xEF && cursor[1] == 0xBB && cursor[2] == 0xB)
+		cursor += 3;
+
+	int lineno = 0;
+	while (*cursor)
+	{
+		lineno++;
+
+		// skip leading whitespace
+		while (q_isblank(*cursor))
+			++cursor;
+
+		char *line = cursor;
+		char *equals = NULL;
+		// find line end and first equals sign, if any
+		while (*cursor && *cursor != '\n')
+		{
+			if (*cursor == '=' && !equals)
+				equals = cursor;
+			cursor++;
+		}
+
+		if (equals)
+		{
+			char *key_end = equals;
+			// trim whitespace before equals sign
+			while (key_end != line && q_isspace(key_end[-1]))
+				key_end--;
+			*key_end = 0;
+
+			char *value = equals + 1;
+			// skip whitespace after equals sign
+			while (value != cursor && q_isspace(*value))
+				value++;
+
+			qboolean leading_quote = (*value == '\"');
+			qboolean trailing_quote = false;
+			value += leading_quote;
+
+			// transform escape sequences in-place
+			char *value_src = value;
+			char *value_dst = value;
+			while (value_src != cursor)
+			{
+				if (*value_src == '\\' && value_src + 1 != cursor)
+				{
+					char c = value_src[1];
+					value_src += 2;
+					switch (c)
+					{
+						case 'n': *value_dst++ = '\n'; break;
+						case 't': *value_dst++ = '\t'; break;
+						case 'v': *value_dst++ = '\v'; break;
+						case 'b': *value_dst++ = '\b'; break;
+						case 'f': *value_dst++ = '\f'; break;
+
+						case '"':
+						case '\'':
+							*value_dst++ = c;
+							break;
+
+						default:
+							Con_Printf("LOC_LoadFile: unrecognized escape sequence \\%c on line %d\n", c, lineno);
+							*value_dst++ = c;
+							break;
+					}
+					continue;
+				}
+
+				if (*value_src == '\"')
+				{
+					trailing_quote = true;
+					*value_dst = 0;
+					break;
+				}
+
+				*value_dst++ = *value_src++;
+			}
+
+			// if not a quoted string, trim trailing whitespace
+			if (!trailing_quote)
+			{
+				while (value_dst != value && q_isblank(value_dst[-1]))
+				{
+					*value_dst = 0;
+					value_dst--;
+				}
+			}
+
+			if (localization.numentries == localization.maxnumentries)
+			{
+				// grow by 50%
+				localization.maxnumentries += localization.maxnumentries >> 1;
+				localization.maxnumentries = q_max(localization.maxnumentries, 32);
+				localization.entries = (locentry_t*) realloc(localization.entries, sizeof(*localization.entries) * localization.maxnumentries);
+			}
+
+			locentry_t *entry = &localization.entries[localization.numentries++];
+			entry->key = line;
+			entry->value = value;
+		}
+
+		if (*cursor)
+			*cursor++ = 0; // terminate line and advance to next
+	}
+
+	// hash all entries
+
+	localization.numindices = localization.numentries * 2; // 50% load factor
+	if (localization.numindices == 0)
+	{
+		Con_DPrintf("No localized strings in file '%s'\n", localization.numentries, path);
+		return;
+	}
+
+	localization.indices = (unsigned*) realloc(localization.indices, localization.numindices * sizeof(*localization.indices));
+	memset(localization.indices, 0, localization.numindices * sizeof(*localization.indices));
+
+	int i;
+	for (i = 0; i < localization.numentries; i++)
+	{
+		locentry_t *entry = &localization.entries[i];
+		unsigned pos = COM_HashString(entry->key) % localization.numindices, end = pos;
+
+		for (;;)
+		{
+			if (!localization.indices[pos])
+			{
+				localization.indices[pos] = i + 1;
+				break;
+			}
+
+			++pos;
+			if (pos == localization.numindices)
+				pos = 0;
+
+			if (pos == end)
+				Sys_Error("COM_LoadLocalization failed");
+		}
+	}
+
+	Con_DPrintf("Loaded %d localized strings from '%s'\n", localization.numentries, path);
+}
+
+/*
+================
+LOC_Init
+================
+*/
+void LOC_Init()
+{
+	LOC_LoadFile("loc_english.txt");
+}
+
+/*
+================
+LOC_GetRawString
+
+Returns localized string if available, or NULL otherwise
+================
+*/
+const char* LOC_GetRawString (const char* key)
+{
+	if (!key || !*key || *key != '$' || !localization.numindices)
+		return NULL;
+	key++;
+
+	unsigned pos = COM_HashString(key) % localization.numindices, end = pos;
+
+	do
+	{
+		unsigned index = localization.indices[pos];
+		if (!index)
+			return NULL;
+
+		locentry_t *entry = &localization.entries[index - 1];
+		if (!Q_strcmp(entry->key, key))
+			return entry->value;
+
+		++pos;
+		if (pos == localization.numindices)
+			pos = 0;
+	} while (pos != end);
+
+	return NULL;
+}
+
+/*
+================
+LOC_GetString
+
+Returns localized string if available, or input string otherwise
+================
+*/
+const char* LOC_GetString (const char* key)
+{
+	const char* value = LOC_GetRawString(key);
+	return value ? value : key;
+}
+
+/*
+================
+LOC_Format
+
+Localizes the format string and performs argument substitution (also localizing the arguments)
+
+Returns number of written chars, excluding the NUL terminator
+If len > 0, output is always NUL-terminated
+================
+*/
+size_t LOC_Format (const char *format, const char* (*getarg) (int index, void* userdata), void* userdata, char* out, size_t len)
+{
+	if (!len)
+	{
+		Con_DPrintf("LOC_Format: no output space\n");
+		return 0;
+	}
+	--len; // reserve space for the terminator
+
+	format = LOC_GetString(format);
+	size_t written = 0;
+	int numargs = 0;
+
+	while (*format && written < len)
+	{
+		if (*format != '{')
+		{
+			out[written++] = *format++;
+			continue;
+		}
+
+		format++;
+		numargs++;
+
+		int argindex = 0;
+		int consumed = 0;
+		sscanf(format, "%d}%n", &argindex, &consumed);
+		format += consumed;
+
+		const char* insert = LOC_GetString(getarg(argindex, userdata));
+		size_t space_left = len - written;
+		size_t insert_len = Q_strlen(insert);
+
+		if (insert_len > space_left)
+		{
+			Con_DPrintf("LOC_Format: overflow at argument #%d\n", numargs);
+			insert_len = space_left;
+		}
+
+		Q_memcpy(out + written, insert, insert_len);
+		written += insert_len;
+	}
+
+	if (*format)
+		Con_DPrintf("LOC_Format: overflow\n");
+
+	out[written] = 0;
+
+	return written;
+}
--- a/Quake/common.c
+++ b/Quake/common.c
@@ -3085,7 +3085,7 @@ LOC_GetRawString
 Returns localized string if available, or NULL otherwise
 ================
 */
-const char* LOC_GetRawString (const char* key)
+const char* LOC_GetRawString (const char *key)
 {
 	unsigned pos, end;
 
@@ -3122,7 +3122,7 @@ LOC_GetString
 Returns localized string if available, or input string otherwise
 ================
 */
-const char* LOC_GetString (const char* key)
+const char* LOC_GetString (const char *key)
 {
 	const char* value = LOC_GetRawString(key);
 	return value ? value : key;
@@ -3130,9 +3130,57 @@ const char* LOC_GetString (const char* key)
 
 /*
 ================
+LOC_ParseArg
+
+Returns argument index (>= 0) and advances the string if it starts with a placeholder ({} or {N}),
+otherwise returns a negative value and leaves the pointer unchanged
+================
+*/
+static int LOC_ParseArg (const char **pstr)
+{
+	int arg;
+	const char *start;
+	const char *str = *pstr;
+
+	// opening brace
+	if (*str != '{')
+		return -1;
+	start = ++str;
+
+	// optional index, defaulting to 0
+	arg = 0;
+	while (q_isdigit(*str))
+		arg = arg * 10 + *str++ - '0';
+
+	// closing brace
+	if (*str != '}')
+		return -1;
+	*pstr = ++str;
+
+	return arg;
+}
+
+/*
+================
+LOC_HasPlaceholders
+================
+*/
+qboolean LOC_HasPlaceholders (const char *str)
+{
+	while (*str)
+	{
+		if (LOC_ParseArg(&str) >= 0)
+			return true;
+		str++;
+	}
+	return false;
+}
+
+/*
+================
 LOC_Format
 
-Localizes the format string and performs argument substitution (also localizing the arguments)
+Replaces placeholders (of the form {} or {N}) with the corresponding arguments
 
 Returns number of written chars, excluding the NUL terminator
 If len > 0, output is always NUL-terminated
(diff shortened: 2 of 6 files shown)


Quake/common.c
```diff
@@ -2839,3 +2840,337 @@ long FS_filelength (fshandle_t *fh)
...
+		if (*cursor)
+			*cursor++ = 0; // terminate line and advance to next
+	}
+
+	// hash all entries
+
+	localization.numindices = localization.numentries * 2; // 50% load factor
+	if (localization.numindices == 0)
+	{
+		Con_DPrintf("No localized strings in file '%s'\n", localization.numentries, path);
```
sezero, line 3028: Lose `localization.numentries, ` here

sezero:
Haven't tried in vkquake, but tried adapting to qs - see attached patch:

- renames `MOVETYPE_EXT_BOUNCEMISSILE` to `MOVETYPE_GIB`
- fixes the unwanted format argument in LOC code I noted
- removes C99'isms
- changes loc file name to `localization/loc_english.txt`

The `illegible server message` and `bad movetype` errors are gone.
However the messages don't seem to be retrieved from the localization file,
i.e. I see `$map_skill_easy`, `$map_start_e1`, `$map_jump_across` and
so on - maybe I made a mistake somewhere when backporting. (Also note
that the localize file is in a zip archive, i.e. no way of extracting it from within
the engine without specifically adding code, which I don't think is worth the
trouble.)

sezero:
The qs version of your patch:  [patch_qs.txt](https://github.com/Novum/vkQuake/files/7027332/patch_qs.txt)

andrei-drexler:
Thank you!

I think I've implemented all your changes, except for one that I wasn't sure about: I've noticed that the [overflow warning](https://github.com/andrei-drexler/vkQuake/blob/ac0aac55b964b9d74422704cc7b8b38ffe3a006a/Quake/pr_cmds.c#L85-L92) in `PF_VarString` was moved to the legacy concatenation code path (the `else` block of `if (localized)`), whereas in my code it was triggered in both cases, since I figured we're generating a large string that would have overflowed WinQuake's buffer either way. It's a minor point, but I wanted to confirm it's intentional.

I've also tested the QS patch on a fresh 0.93.3 codebase, with the localization file copied to both `mg1/localization` and `id2/localization` (with `id2` being a copy of `id1` from the re-release), and the strings were properly localized in both cases. Note that each mod gets its own localization - the right choice for an open engine, IMO, but different from the re-release - maybe that's the issue?

As for the zip support, I completely agree, and I tried to keep this patch as simple as possible. For true localization we'd also have to add UTF-8 support, UI integration, and maybe per-map localization files, but those are different issues for another time.

sezero:
>  I've noticed that the [overflow warning](https://github.com/andrei-drexler/vkQuake/blob/ac0aac55b964b9d74422704cc7b8b38ffe3a006a/Quake/pr_cmds.c#L85-L92) in `PF_VarString` was moved to the legacy concatenation code path (the `else` block of `if (localized)`), whereas in my code it was triggered in both cases, since I figured we're generating a large string that would have overflowed WinQuake's buffer either way. It's a minor point, but I wanted to confirm it's intentional.

Not intentional, I missed it I guess. BTW, why did you changed the error into a warning? (haven't yet tested)

> I've also tested the QS patch on a fresh 0.93.3 codebase, with the localization file copied to both `mg1/localization` and `id2/localization` (with `id2` being a copy of `id1` from the re-release), and the strings were properly localized in both cases. Note that each mod gets its own localization - the right choice for an open engine, IMO, but different from the re-release - maybe that's the issue?

Ah, I totally missed the quakefs behavior to load from gamedir, not basedir: moving the localization file under id1 works.

> As for the zip support, I completely agree, and I tried to keep this patch as simple as possible. For true localization we'd also have to add UTF-8 support, UI integration, and maybe per-map localization files, but those are different issues for another time.

Yes: That alone makes the whole localization code here non-functional: you want to add it for reference purposes?

Quake/cl_parse.c
```diff
@@ -1822,9 +1822,11 @@ void CL_ParseServerMessage (void)
 		case svc_updatefrags:
 			Sbar_Changed ();
 			i = MSG_ReadByte ();
-			if (i >= cl.maxclients)
-				Host_Error ("CL_ParseServerMessage: svc_updatefrags > MAX_SCOREBOARD");
-			cl.scores[i].frags = MSG_ReadShort ();
+			j = MSG_ReadShort ();
+			if (i < 0 || i >= cl.maxclients)
+				Con_DPrintf ("CL_ParseServerMessage: svc_updatefrags > MAX_SCOREBOARD\n");
```
sezero, line 1827: Why was this necessary? Do we hit this error in the r-release version?

andrei-drexler: Was hitting this when loading mg1/id2 savegames (receiving a `svc_updatefrags` when `cl.maxclients` was still 0), but it only happens when using the fte666/fte999 protocols, so QS doesn't need it. Strangely enough, setting `cl_nopext 1` doesn't fix it.

andrei-drexler:
> moving the localization file under id1 works.

:thumbsup:

> Yes: That alone makes the whole localization code here non-functional: you want to add it for reference purposes?

I wanted to add this mostly for compatibility with the re-release (with only a small bit of manual user intervention), but I think it can also serve as a good first step towards true localization. Even without UTF-8, basic support for Italian, French, Spanish, German, latinized Russian etc. could be added relatively easily in the future. It wouldn't be as comprehensive as what the NightDive engine offers, but it also wouldn't be a whole lot of extra code.

Quake/cl_parse.c
sezero: I tried running mg1 with the minimal three changes in qs, saved and loaded,
got no errors: is the error random or guaranteed to happen?

sezero:
Pushed the three minimal changes with your name. Please rebase.

Quake/cl_parse.c
andrei-drexler: In vkQuake it happens every time, but I haven't seen it at all in QS, so it's probably safe to remove this change from the QS patch.

sezero: BTW, applied the same three changes in here to qs too.

sezero: > In vkQuake it happens every time, but I haven't seen it at all in QS, so it's probably safe to remove this change from the QS patch.

OK then.

sezero: > In vkQuake it happens every time, but I haven't seen it at all in QS, 

I actually wonder why it is happening in vkquake - that you are saying it still happens with `cl_nopext 1`, curious

sezero:
Wtih the three minimal parts in, which avoids the errors and the `campaign` spam,
I have reservations about the rest, as explained above. To summarize:

- Changing the `Host_Error` to simply a warning in `CL_ParseServerMessage`::
  `svc_updatefrags` case may not be a good idea. Loading saves in QS doesn't
  error, so we should find the reason for errors in vkQuake.

- Localized strings support doesn't work unless the user manually extracts the
  relevant files to under `id1/`. Therefore we need not those parts, save for them
  being for future reference.

Leaving the decisions to @Novum

sezero:
BTW, someone posted a link to achievement thing on sf.net/qs site; for reference,
in case anyone is interested: https://steamdb.info/app/2310/stats/

Quake/cl_parse.c
andrei-drexler: I've looked into this a bit more, and it seems that when the FTE protocol extensions are active, `SV_SendServerinfo` will first ask the client what extensions it supports
https://github.com/Novum/vkQuake/blob/18df3c1cdcfea8da4a33750cd781ab9a30eed1f7/Quake/sv_main.c#L1289-L1295
and only then send the actual server info message (`SV_Pext_f`)
https://github.com/Novum/vkQuake/blob/18df3c1cdcfea8da4a33750cd781ab9a30eed1f7/Quake/sv_main.c#L1512-L1513
The problem is that on the same frame that the client connects and the extension query is sent, the server will also send `svc_updatefrags` messages (from `SV_UpdateToReliableMessages`, called from `SV_SendClientMessages`)
https://github.com/Novum/vkQuake/blob/18df3c1cdcfea8da4a33750cd781ab9a30eed1f7/Quake/sv_main.c#L2263-L2279
The `svc_updatefrags` arrives at the client before `svc_serverinfo`, which means that `cl.maxclients` is still 0, hence the error.

A possible solution (other than turning the error into a warning) might be to change
https://github.com/Novum/vkQuake/blob/18df3c1cdcfea8da4a33750cd781ab9a30eed1f7/Quake/sv_main.c#L2270
to
```c
if (!client->active || client->sendsignon == PRESPAWN_FLUSH)
```
to avoid sending `svc_updatefrags` before `svc_serverinfo`. It seems to work, but I've only done some limited testing and I'm not exactly sure of the implications, especially for multiplayer mode. I'd say that changing the error to a warning still seems like the safest solution, but I'm also wondering why this isn't triggered with any other mod.

sezero: @Shpoike: comments?

vide0hanz:
I'm currently testing this but cannot get the localization strings to appear correctly. I have tried putting `loc_english.txt` in id1/ as well as `id1/localization/loc_english.txt`. The spam campaign error messages are resolved, and the game no longer crashes when triggering an achievement; only issue for me is the localization. 

If I understand the comments above, extracting `loc_english.txt` into id1 should work, correct?

bangstk:
>     * Localized strings support doesn't work unless the user manually extracts the
>       relevant files to under `id1/`. Therefore we need not those parts, save for them
>       being for future reference.
I think this should be kept in, because without this, all the centerprint messages show up as stuff like "$map_skill_hard" ingame. Yes it does take a bit more effort for an end user (extracting the strings file from the QuakeEx.kpf, which is just a renamed zip file), but it's preferable to not having readable strings at all. Unless you would like to hardcode the English strings directly into the engine....

vide0hanz:
@andrei-drexler Do you happen to have a patch file handy for all of your proposed changes here?

andrei-drexler:
> @andrei-drexler Do you happen to have a patch file handy for all of your proposed changes here?

Sure, [here it is](https://github.com/Novum/vkQuake/pull/345.patch) (a neat trick that I only learned recently - you can append `.patch` or `.diff` to the URL of a pull request or commit to get a plain-text version).

vide0hanz:
> > @andrei-drexler Do you happen to have a patch file handy for all of your proposed changes here?
> 
> Sure, [here it is](https://github.com/Novum/vkQuake/pull/345.patch) (a neat trick that I only learned recently - you can append `.patch` or `.diff` to the URL of a pull request or commit to get a plain-text version).

Oh nice, I didn't know that. Thanks! Just wanted this for reference.

vide0hanz:
@andrei-drexler Hmm, while the localization stuff works, for some reason the messages which are displayed when completing an episode still show their variable names:

`$qc_finale_e1` for example.

andrei-drexler:
@sezero: 
> I have reservations about the rest, as explained above. To summarize:
>
>     * Changing the `Host_Error` to simply a warning in `CL_ParseServerMessage`::
>       `svc_updatefrags` case may not be a good idea. Loading saves in QS doesn't
>       error, so we should find the reason for errors in vkQuake.
>
>     * Localized strings support doesn't work unless the user manually extracts the
>       relevant files to under `id1/`. Therefore we need not those parts, save for them
>       being for future reference.
> 

For the `svc_updatefrags` error, it's true that I'm only treating a symptom instead of addressing the root cause, but this fix seemed a lot safer than the alternative, and it can easily be reverted once we have a better solution. In the meantime, though, the new content isn't really playable if loading a savefile doesn't work.

Setting aside the savegame issue (which QS doesn't have anyway), even if the maps are *technically* playable, they're still less *enjoyable* IMO because of the missing localization - seeing `$map_mg1_intro_e1` doesn't have the same effect as `This is the realm of Machinists. The cogs are turning`. For the kind of players that would even consider trying the update in QuakeSpasm or vkQuake, I also think unzipping the localization files manually shouldn't be much of a problem if the process is clearly documented in the readme/release notes, and seems preferable to having illegible message throughout the whole game.

sezero:
> For the `svc_updatefrags` error, it's true that I'm only treating a symptom instead of addressing the root cause, but this fix seemed a lot safer than the alternative, and it can easily be reverted once we have a better solution. In the meantime, though, the new content isn't really playable if loading a savefile doesn't work.
> 
> Setting aside the savegame issue (which QS doesn't have anyway),

I wonder whether QSS has the issue

>  even if the maps are _technically_ playable, they're still less _enjoyable_ IMO because of the missing localization - seeing `$map_mg1_intro_e1` doesn't have the same effect as `This is the realm of Machinists. The cogs are turning`. For the kind of players that would even consider trying the update in QuakeSpasm or vkQuake, I also think unzipping the localization files manually shouldn't be much of a problem if the process is clearly documented in the readme/release notes, and seems preferable to having illegible message throughout the whole game.

I see your points, I do.  If we are to apply the localization though, we
- either have to document what the users have to do,
- or add some unzip code to extract the localization file

Either way, it still is problematic when e.g. you have a network play:
If a rerelease server sends the localized message, the clients shall
have no means at all to decode it unless they also are running the
rerelease thing (no, I haven't tested that, just theorizing.)

I still wonder what @Novum has to say on the matter?

vide0hanz:
I would **think** that the localization stuff is mostly for client-side messages. That said, even with this current approach, it is hard coded to look for the loc_english.txt file. Not sure if it can be made more generic to just look for `*.txt` instead, otherwise adding in ` || loc_french.txt || loc_german.txt || loc_italian.txt || ...etc` would be needed, which also doesn't seem ideal. 

While I am definitely in favor of having something which works as a stop-gap just to play the rerelease content, I have to wonder if there is a better way of going about it.

andrei-drexler:
@vide0hanz:
> 
> 
> @andrei-drexler Hmm, while the localization stuff works, for some reason the messages which are displayed when completing an episode still show their variable names:
> 
> `$qc_finale_e1` for example.

It should work, I've just tested on e1m7 (and also in the new campaign earlier)

vide0hanz:
@andrei-drexler Odd. I have no idea why these specific variables aren't localizing on my end when everything else works fine. I'm using a new pakfile I created which includes all of the updated models, and while I can't think of any obvious reason why that would affect this, its possible I'm missing something there.

andrei-drexler:
> This error is still curious though, happens as soon as you beat the base game:

Good catch, fixed in 787ab3f020bb673dd0151cd77e0fdfdcf9e7bab0

sezero:
We also need some dummy `menu_credits` command. The following
is e.g. from dopa or mg1 when you complete the game.
```
             WE SALUTE YOU.

Unknown command "menu_credits"
Sending clc_disconnect
Client player removed
```

andrei-drexler:
> We also need some dummy `menu_credits` command.

Oh, you're right, I saw that too a couple of times, but I was always in the middle of something else and never got to it. https://github.com/Novum/vkQuake/commit/2fd899336fc797b1f74ac35b1e5fed1a7a39b3ae

Quake/cl_parse.c
andrei-drexler: I tried QSS, and it doesn't crash when loading. The diffference is that it's [checking](https://github.com/Shpoike/Quakespasm/blob/c48936385a4428ce701059a11215780ec3cdcc00/quakespasm/Quake/sv_main.c#L2916) `client->knowntoqc` instead of `client->active` - I'll commit this and rebase.

sezero: OK

sezero:
Hit this in hipnotic:
```
You got Mjolnir
Gremlin stole your Rocket Launcher
You got {}$qc_rocket_launcher, 4 rockets
You receive 25 health
```

sezero:
> Hit this in hipnotic:
> 
> ```
> You got Mjolnir
> Gremlin stole your Rocket Launcher
> You got {}$qc_rocket_launcher, 4 rockets
> You receive 25 health
> ```

Seems to always happen when gremlins steal your weapon
and then you kill it and take it back:
```
Gremlin stole your Rocket Launcher
You got {}$qc_rocket_launcher, 5 rockets
Gremlin stole your Lightning Gun
You got {}$qc_thunderbolt, 38 cells
Gremlin stole your Nailgun
You got {}$qc_nailgun, 36 nails
Gremlin stole your Lightning Gun
You got {}$qc_thunderbolt, 40 cells
```

andrei-drexler:
Hmm, it looks like some of my assumptions about code formatting were wrong, I'll have to handle that a bit differently. Thanks for the save file!

sezero:
The latest commit resists gremlins, so far :)

andrei-drexler:
Yup, that's it!

sezero:
Pushed commit 1c9d6bdafc for it to this branch.

andrei-drexler:
I think I have a solution for the ents problem, too: when trying to load the external file, start by computing the CRC of the embedded entity lump. Let's say we're loading `e1m1`, and its CRC is `0x04c8`. First, we try `maps/e1m1@04c8.ent`. If that isn't found, fall back to `maps/e1m1.ent`. If that's not found either, load the original data. This way, we could rename the existing ents in `quakespasm.pak` and not have to worry about interference.

Something like this:
```diff
diff --git a/Quake/gl_model.c b/Quake/gl_model.c
index b7060a2..bb3f3ef 100644
--- a/Quake/gl_model.c
+++ b/Quake/gl_model.c
@@ -798,20 +798,31 @@ Mod_LoadEntities
 */
 void Mod_LoadEntities (lump_t *l)
 {
-	char	entfilename[MAX_QPATH];
+	char	basemapname[MAX_QPATH], entfilename[MAX_QPATH];
 	char		*ents;
 	int		mark;
-	unsigned int	path_id;
+	unsigned int	path_id, crc;
 
 	if (! external_ents.value)
 		goto _load_embedded;
 
-	q_strlcpy(entfilename, loadmodel->name, sizeof(entfilename));
-	COM_StripExtension(entfilename, entfilename, sizeof(entfilename));
-	q_strlcat(entfilename, ".ent", sizeof(entfilename));
-	Con_DPrintf2("trying to load %s\n", entfilename);
+	crc = CRC_Block(mod_base + l->fileofs, l->filelen);
 	mark = Hunk_LowMark();
+
+	q_strlcpy(basemapname, loadmodel->name, sizeof(basemapname));
+	COM_StripExtension(basemapname, basemapname, sizeof(basemapname));
+
+	q_snprintf(entfilename, sizeof(entfilename), "%s@%04x.ent", basemapname, crc);
+	Con_DPrintf2("trying to load %s\n", entfilename);
 	ents = (char *) COM_LoadHunkFile (entfilename, &path_id);
+
+	if (!ents)
+	{
+		q_snprintf(entfilename, sizeof(entfilename), "%s.ent", basemapname);
+		Con_DPrintf2("trying to load %s\n", entfilename);
+		ents = (char *) COM_LoadHunkFile (entfilename, &path_id);
+	}
+
 	if (ents)
 	{
 		// use ent file only from the same gamedir as the map
```

sezero:
> I think I have a solution for the ents problem, too: when trying to load the external file, start by computing the CRC of the embedded entity lump. Let's say we're loading `e1m1`, and its CRC is `0x04c8`. First, we try `maps/e1m1@04c8.ent`. If that isn't found, fall back to `maps/e1m1.ent`. If that's not found either, load the original data. This way, we could rename the existing ents in `quakespasm.pak` and not have to worry about interference.
> 
> Something like this:
> 
> ```diff
> diff --git a/Quake/gl_model.c b/Quake/gl_model.c
> index b7060a2..bb3f3ef 100644
> --- a/Quake/gl_model.c
> +++ b/Quake/gl_model.c
> @@ -798,20 +798,31 @@ Mod_LoadEntities
>  */
>  void Mod_LoadEntities (lump_t *l)
>  {
> -	char	entfilename[MAX_QPATH];
> +	char	basemapname[MAX_QPATH], entfilename[MAX_QPATH];
>  	char		*ents;
>  	int		mark;
> -	unsigned int	path_id;
> +	unsigned int	path_id, crc;
>  
>  	if (! external_ents.value)
>  		goto _load_embedded;
>  
> -	q_strlcpy(entfilename, loadmodel->name, sizeof(entfilename));
> -	COM_StripExtension(entfilename, entfilename, sizeof(entfilename));
> -	q_strlcat(entfilename, ".ent", sizeof(entfilename));
> -	Con_DPrintf2("trying to load %s\n", entfilename);
> +	crc = CRC_Block(mod_base + l->fileofs, l->filelen);
>  	mark = Hunk_LowMark();
> +
> +	q_strlcpy(basemapname, loadmodel->name, sizeof(basemapname));
> +	COM_StripExtension(basemapname, basemapname, sizeof(basemapname));
> +
> +	q_snprintf(entfilename, sizeof(entfilename), "%s@%04x.ent", basemapname, crc);
> +	Con_DPrintf2("trying to load %s\n", entfilename);
>  	ents = (char *) COM_LoadHunkFile (entfilename, &path_id);
> +
> +	if (!ents)
> +	{
> +		q_snprintf(entfilename, sizeof(entfilename), "%s.ent", basemapname);
> +		Con_DPrintf2("trying to load %s\n", entfilename);
> +		ents = (char *) COM_LoadHunkFile (entfilename, &path_id);
> +	}
> +
>  	if (ents)
>  	{
>  		// use ent file only from the same gamedir as the map
> ```

@ericwa: What do you say?

ericwa:
Yeah, that sounds like an improvement (renaming the .ent patches in `quakespasm.pak` to e.g. `maps/e1m1@04c8.ent`).

The .ent file patches as a solution to z-fighting in the id1 maps has never been to most elegant, but having a CRC check would make it safer at least.

I checked if FTEQW does anything like that, and it doesn't (just regular "mapname.ent" support.)

sezero:
OK, will get back to the ent patch support tomorrow (I hope.)

Andrei: with the ent patch, we can remove the default.cfg patch and
instruct users to specifically copy quakespasm.pak to rerelease dir:
yes?

andrei-drexler:
I'd still keep the default cfg in the exe, it just seems safer if it can function on its own. This way the pak is just a bonus, and since it's not needed in the rerelease, I wouldn't mention it at all in the instructions, just to keep them as short as possible:

> vkQuake has initial support for playing the 2021 re-release content:
> - Copy the vkQuake exe to your rerelease installation
> - Extract the localization folder from the QuakeEX.kpf zip file and place it in the rerelease directory
> - Run vkQuake as you normally do

Maybe we should also say something about using `game mg1` to play DOTM, since there's no way of doing it via the in-game UI?

As for the code itself, if the gl_model.c diff above looks ok, I can commit it along with the renamed ent files and the updated pak.

sezero:
> I'd still keep the default cfg in the exe, it just seems safer if it can function on its own. This way the pak is just a bonus, and since it's not needed in the rerelease, I wouldn't mention it at all in the instructions, just to keep them as short as possible:

OK

> > vkQuake has initial support for playing the 2021 re-release content:
> > 
> > * Copy the vkQuake exe to your rerelease installation
> > * Extract the localization folder from the QuakeEX.kpf zip file and place it in the rerelease directory
> > * Run vkQuake as you normally do
> 
> Maybe we should also say something about using `game mg1` to play DOTM, since there's no way of doing it via the in-game UI?

Hm?? There is no way of playing any other mission packs etc using the in-game UI either, what is special about mg1?

> As for the code itself, if the gl_model.c diff above looks ok, I can commit it along with the renamed ent files and the updated pak.

OK

andrei-drexler:
Rebased (since there were changes to vkquake.pak on master) and pushed https://github.com/Novum/vkQuake/pull/345/commits/79f5de5e682aa693f182db45d54db838ec6986b4. It looks like the old .ent.orig files in the repo had newlines added at the end, the versions I committed don't, they're straight from the bsp's (since we're crc-ing them, after all).

> Hm?? There is no way of playing any other mission packs etc using the in-game UI either, what is special about mg1?

Just trying to make it easier for new players whose first contact with Quake was the re-release (which *does* show the new expansions in the UI) and might not know about the game command.

sezero:
>  It looks like the old .ent.orig files in the repo had newlines added at the end, the versions I committed don't, they're straight from the bsp's (since we're crc-ing them, after all).

In the new commit, they have a nul-terminator at the end instead of newline:
How did you extract the ents into files? Do the extracted files crcs match the
in-game generated ones?

> 
> > Hm?? There is no way of playing any other mission packs etc using the in-game UI either, what is special about mg1?
> 
> Just trying to make it easier for new players whose first contact with Quake was the re-release (which _does_ show the new expansions in the UI) and might not know about the game command.

Well, if someone is wanting to play rerelease using qs or vkquake, it means they either can't run the rerelease as is and/or don't care about rerelease's new features, either way mentioning that is unnecessary: qs has documentation for the game command (feel free to mention it in vk vrsion if you want, though.)

sezero:
> > It looks like the old .ent.orig files in the repo had newlines added at the end, the versions I committed don't, they're straight from the bsp's (since we're crc-ing them, after all).
> 
> In the new commit, they have a nul-terminator at the end instead of newline:
> (correction: '\0' _after_ the last newline)
> How did you extract the ents into files? Do the extracted files crcs match the
> in-game generated ones?

OK, the ents in the pak do contain a nul terminator and the crcs orig files that you
replaced do match the in-game generated ones. I don't like the nul-terminator though.

andrei-drexler:
> In the new commit, they have a nul-terminator at the end instead of newline:
> (correction: '\0' _after_ the last newline)
> How did you extract the ents into files? Do the extracted files crcs match the
> in-game generated ones?
> 

I've used `COM_WriteFile` during loading, passing it entire the lump data. The zeroes are present in the BSP, but I guess I could check for a terminating NUL, and only hash (and re-export) the preceding part. It seems more hacky/error-prone, but it will clean up the .orig file.

sezero:
>  I could check for a terminating NUL, and only hash (and re-export) the preceding part. It seems more hacky/error-prone, but it will clean up the .orig file.

No need to add extra complexity I guess.

andrei-drexler:
Hmm. I have mixed feelings about this, but I think it would be a good idea to be able to CRC .ent files extracted with bsputil and have them match if nothing's changed. And it seems [bsputil](https://github.com/ericwa/ericw-tools/blob/8beda9a03918dcd217818f428a9175c3e6bfbd03/bsputil/bsputil.cc#L604) doesn't include the NUL.

sezero:
Well, if the only change will be `l->filelen - 1` then it's OK

andrei-drexler:
Unfortunately, it seems bsputil writes files in [text mode](https://github.com/ericwa/ericw-tools/blob/8beda9a03918dcd217818f428a9175c3e6bfbd03/bsputil/bsputil.cc#L600), so there will be newline differences anyway...

sezero:
> Unfortunately, it seems bsputil writes files in [text mode](https://github.com/ericwa/ericw-tools/blob/8beda9a03918dcd217818f428a9175c3e6bfbd03/bsputil/bsputil.cc#L600), so there will be newline differences anyway...

Ugh. That can happen on windoze.. just dos2unix the extracted files.

(BTW, your generated pak has all the files in dos cr/lf mode..)

sezero:
OK, I think the following should be good, and we can go back to old orig files,
If it looks good to you feel free to apply and update crcs.
```diff
diff --git a/Quake/gl_model.c b/Quake/gl_model.c
index d15bebf..f8ed30c 100644
--- a/Quake/gl_model.c
+++ b/Quake/gl_model.c
@@ -801,12 +801,14 @@ void Mod_LoadEntities (lump_t *l)
 	char	basemapname[MAX_QPATH], entfilename[MAX_QPATH];
 	char		*ents;
 	int		mark;
-	unsigned int	path_id, crc;
+	unsigned int	path_id;
+	unsigned int	crc = 0;
 
 	if (! external_ents.value)
 		goto _load_embedded;
 
-	crc = CRC_Block(mod_base + l->fileofs, l->filelen);
+	if (l->filelen > 0)
+		crc = CRC_Block(mod_base + l->fileofs, l->filelen - 1);
 	mark = Hunk_LowMark();
 
 	q_strlcpy(basemapname, loadmodel->name, sizeof(basemapname));
```

andrei-drexler:
> Ugh. That can happen on windoze..

And it did, that's how I found the issue :(

> just dos2unix the extracted files

Sure, just that I was hoping other users could extract a crc-matching file with bsputil without any manual intervention, but it seems manual intervention is a recurring theme...

> (BTW, your generated pak has all the files in dos cr/lf mode..)

Ouch, I was wondering why they were bigger... I guess that's what `mkpak.sh` ends up doing on windows (using msys). It looks like I'll have to reboot try this again using ubuntu.

> OK, I think the following should be good, and we can go back to old orig files,
If it looks good to you feel free to apply and update crcs.

Alright.

sezero:
> Ouch, I was wondering why they were bigger... I guess that's what `mkpak.sh` ends up doing on windows (using msys). It looks like I'll have to reboot try this again using ubuntu.

Don't worry about that, you finalize everything else, I will regenerate the pak
And we can then start merging: (1) ent-crc patch, (2) default.cfg patch,
and finally (3) localization patch.

sezero:
To verify each other's results, with the nul-terminator removed the crcs I found are:
e1m1 -> c49d
e1m2 -> 0caa
e1m4 -> 958e
e2m2 -> fbfe
e2m3 -> 237a
e2m7 -> 10a8

andrei-drexler:
> Don't worry about that, you finalize everything else, I will regenerate the pak
And we can then start merging: (1) ent-crc patch, (2) default.cfg patch,
and finally (3) localization patch.

If you're already taking care of the ents/paks (thank you, I've rebooted but it looks like you're already ahead), I'm not sure what's left to do besides the readme tweaks :)

> To verify each other's results, with the nul-terminator removed the crcs I found are:
> e1m1 -> c49d
> e1m2 -> 0caa
> e1m4 -> 958e
> e2m2 -> fbfe
> e2m3 -> 237a
> e2m7 -> 10a8

Confirmed.

sezero:
> If you're already taking care of the ents/paks (thank you, I've rebooted but it looks like you're already ahead), I'm not sure what's left to do besides the readme tweaks :)

Pushed f9156f10ff to git master: make sure all are OK (and that ents are identified correctly at runtime)

sezero:
And pushed the default.cfg patch as cbd25e62f422

andrei-drexler:
The - 1 was missing when CRC-ing, but with it in place all the ent files load correctly. I also force-pushed the rerelease branch back to the commit before the default_cfg change.

sezero:
Merged the rest, i.e. the localization stuff.

Let's do any further improvement in master (or with new pull requests if necessary.)

andrei-drexler:
Alright, thank you helping out with all this, and thanks to everyone else that tested the in-progress code and offered feedback.

sezero:
Now pushed the patch to QS, too.

andrei-drexler:
I've only looked at this very briefly earlier, but there seemed to be significant differences when I compared the entity list from a rerelease map to the old version - at the very least the order of the entities has changed, so I'm not sure we can safely take the new fog/waterlevel values and apply them to the old entity files to produce a new pak. Plus, I really think z-fighting should be dealt with by the renderer - after all, the same maps with the same problematic entities don't have this issue in the new engine.

PR #350, opened 2021-08-29
Change gl_zfix to use depth bias and enable it by default

andrei-drexler:
This addresses the z-fighting that is still present in the remastered maps due to co-planar geometry, but without requiring manual entity replacements. This change does introduce some artifacts around the edges of bmodels that are aligned with the surrounding map geometry (e.g. the SSG door in e1m1), but unlike the old gl_zfix implementation that shifted entities around, the artifacts are much more subtle. This is also consistent with the new re-release engine - no z-fighting, but some minor artifacts around hidden doors.

### DIFF
--- a/Quake/gl_rmisc.c
+++ b/Quake/gl_rmisc.c
@@ -1467,7 +1467,7 @@ R_CreatePipelines
 void R_CreatePipelines()
 {
 	int render_pass;
-	int alpha_blend, alpha_test, fullbright_enabled, use_zbias;
+	int alpha_blend, alpha_test, fullbright_enabled;
 	VkResult err;
 
 	Sys_Printf("Creating pipelines\n");
@@ -1942,8 +1942,9 @@ void R_CreatePipelines()
 	rasterization_state_create_info.polygonMode = VK_POLYGON_MODE_FILL;
 	depth_stencil_state_create_info.depthTestEnable = VK_TRUE;
 	depth_stencil_state_create_info.depthWriteEnable = VK_TRUE;
-	rasterization_state_create_info.depthBiasEnable = VK_FALSE;
+	rasterization_state_create_info.depthBiasEnable = VK_TRUE;
 	input_assembly_state_create_info.topology = VK_PRIMITIVE_TOPOLOGY_TRIANGLE_LIST;
+	dynamic_states[dynamic_state_create_info.dynamicStateCount++] = VK_DYNAMIC_STATE_DEPTH_BIAS;
 
 	VkVertexInputAttributeDescription world_vertex_input_attribute_descriptions[3];
 	world_vertex_input_attribute_descriptions[0].binding = 0;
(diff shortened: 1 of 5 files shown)


Novum:
I highly recommend using VK_DYNAMIC_STATE_DEPTH_BIAS instead of adding another pipeline permutation.

andrei-drexler:
Done. I've also tested this on both Nvidia (with D24 and D32F zbuffers) and AMD to make sure the bias factors work as intended on both platforms - I tried to keep them as low as possible to minimize the artifacts, but still high enough to fix z-fighting reliably on the maps what were previously handled by ent patches.

Novum:
Also check 16 bit depth maybe. The bias values are quite different for that.

andrei-drexler:
Is 16 bit ever used, though? The code in gl_vidsdl only checks for D24 and D32F.

andrei-drexler:
Added them just in case (tested on AMD).

Novum:
You are right. Vulkan guarantees either D24S8 or D32S8 and I changed the code to pick either one when doing the sky stencil buffer optimization. Sorry my memory is terrible.

https://github.com/Novum/vkQuake/blob/master/Quake/gl_vidsdl.c#L1005

Quake/r_world.c
```diff
@@ -397,6 +397,11 @@ static void R_FlushBatch (qboolean fullbright_enabled, qboolean alpha_test, qboo
 				constant_factor = -0.125f;
 				slope_factor = -0.125f;
 			}
+			else if (vulkan_globals.depth_format == VK_FORMAT_D16_UNORM_S8_UINT || vulkan_globals.depth_format == VK_FORMAT_D16_UNORM)
```
Novum, line 400: Please revert, we don't use 16 bit depth at all anymore. Sorry, that was my bad.

andrei-drexler: No worries, it was just a quick change :) Should I also remove the checks for the depth-only formats?

andrei-drexler:
No problem! Reverted.

Quake/gl_rmisc.c
```diff
@@ -1942,8 +1942,9 @@ void R_CreatePipelines()
 	rasterization_state_create_info.polygonMode = VK_POLYGON_MODE_FILL;
 	depth_stencil_state_create_info.depthTestEnable = VK_TRUE;
 	depth_stencil_state_create_info.depthWriteEnable = VK_TRUE;
-	rasterization_state_create_info.depthBiasEnable = VK_FALSE;
+	rasterization_state_create_info.depthBiasEnable = VK_TRUE;
```
Novum, line 1945: This doesn't actually get used if `VK_DYNAMIC_STATE_DEPTH_BIAS` is included in the dynamic states. Would be better to just completely remove this line to indicate that it's not used.

`VK_DYNAMIC_STATE_DEPTH_BIAS specifies that the depthBiasConstantFactor, depthBiasClamp and depthBiasSlopeFactor states in VkPipelineRasterizationStateCreateInfo will be ignored and must be set dynamically with vkCmdSetDepthBias before any draws are performed with depthBiasEnable in VkPipelineRasterizationStateCreateInfo set to VK_TRUE.`

PR #354, opened 2021-08-31
Read localization from QuakeEX.kpf (libzip)

Novum:
Let me know what you think. @sezero had doubts because of the additional lib, but I think this is pretty minimal.

Not a final patch, x86 libs & makefile stuff are missing.


sezero:
It works. A few notes:

- windows x86 version of libzip is missing.

- I tested in quakespasm using an old version (0.9) which was available
  in my old distro which had the `zip_t`, `zip_file_t` and `zip_stat_t`
  typedefs missing.  Replacing their uses with explicit `struct zip`,
  `struct zip_file` and `struct zip_stat` just worked:
```
	struct zip *za = NULL;
	struct zip_file *zf = NULL;
	struct zip_stat zs;
```

Novum:
Yeah as I said, x86 libs are missing. I'll add them if everyone is okay with the whole thing.

sezero:
Looking at libzip, it's crazy: as an all-purposes zip library,
all those cryptography options may be good but blah.  Can't we
maybe e.g. quake3's zlib/unzip code or something?

Miniz (https://github.com/richgel999/miniz) is another option
(it's another beast, but can be more configurable IIRC, and it
won't require an additional dll.)

andrei-drexler:
[Side note, in case you're wondering what's with the pushes - this code actually helped me find a bug, and I thought fixing it on main (change "r" to "rb" when opening the file) wouldn't cause a merge conflict here, which it did. Long story short, reverted the changes to avoid causing unnecessary friction - rebasing for a single character seemed excessive.]

As for the actual implementation of this feature, I don't have practical experience with either miniz or libzip, but I guess miniz does seem preferable since it's lighter-weight and it doesn't require a separate DLL. I don't have any strong feelings either way, though. When I needed zip support in my own code some years ago I went with parsing the zip structures manually and using zlib for compression/decompression, and if the goal is to avoid adding new libraries, a similar solution would probably be to reuse lodepng's deflate implementation. I don't think it's worth the effort, though, the zip format is somewhat byzantine and just parsing it would need a few hundred lines of code, as opposed to a couple dozen for the whole feature with libzip/miniz.

Whichever way you decide it's best, though, I am very much looking forward to seeing this merged in.

sezero:
If it were up to me, I'd prefer a zlib-based solution: zlib is universally
available on every unix os including mac, its minizip:unzip.c can be
used for this purpose .

Novum:
The zlib library doesn't actually contain minizip though. It's an extra thing and extra package e.g. on Ubuntu

Novum:
@sezero Considering that it's not part of zlib, do you still prefer minizip? Would you want us to just include `unzip.c` as a source file into the vkQuake tree?

No strong opinions on it, just picked libzip because it seemed more actively maintained.

Novum:
I'll just code it with minizip and create another PR so we can have a better discussion.

Novum:
Sorry accidentally closed this

sezero:
Is the following good? It implements the extraction using miniz:

Patch file (includes Andrei's fopen mode fix): [patch_with_miniz.txt](https://github.com/Novum/vkQuake/files/7086446/patch_with_miniz.txt)
Highly stripped down miniz: [stripped_down_miniz.zip](https://github.com/Novum/vkQuake/files/7086447/stripped_down_miniz.zip)

Tested only its equivalent in quakespasm yet using gcc under i686-linux,
adds about 20k to binary size.

andrei-drexler:
Realloc'ing just for the string terminator seems a bit unusual, but other than that it looks good.

sezero:
> Realloc'ing just for the string terminator seems a bit unusual,

Yeah, I couldn't find anything better: really open to ideas.

@Novum: what do you think?

Novum:
@sezero Could you create a PR? Easier to review stuff.

sezero:
> @sezero Could you create a PR? Easier to review stuff.

Done: #357

PR #360, opened 2021-09-03
Resolves #342 - Update cl.scores  properly on connect and disconnect

### LINKED ISSUE #342: In coop, players already on the server when client joins appear as all-white
https://github.com/Shpoike/Quakespasm/issues/43 This QSS bug seems to apply to vkquake (1.10.0-beta3 win64) also, but in a  different way.  In a listen server started in-game, every player who was already on the server when the client joins shows as all-white, but other players who join after the client have correct colors.  I messed around on a dedicated server and this continued to be true, but also if the all-white players re-establish their colors by doing the "color" console command and suiciding once, when they respawn they will have correct colors henceforth, even across map changes.  (Notably, the color command must be used, just suiciding does nothing, and in QSS this trick does not work.)

Perpixel:


Novum:
Does this mirror the QSS code?

Perpixel:
> 
> 
> Does this mirror the QSS code?

Not familiar with what QSS did but its somewhat similar. Not sure why both functions got so far apart from QSS to vkQuake...

andrei-drexler:
I don't think these changes break the protocol - they're not adding any new message types and they're not changing the wire format of existing ones, either. They're also server side only, which means existing clients can already handle this message sequence.

As for matching QSS, since QSS has a similar issue (it might actually be identical now after https://github.com/Novum/vkQuake/commit/9a8a95e933ca0df21530c255ac7561ca05bfd59c), diverging from it is necessary to fix this.

Perpixel:
This is just 2 extra messages you could be getting at any other time in a match anyway. At the end of the day we really don't have a choice but to send both message when someone join or leave to keep player.scores updated. I can cross check QSS a bit but I really don't think this ever worked correctly since "optimized" or even it what scenario it can work.

Also I'm not sure how the entity delta was supposed to handle this but I'll look.

Perpixel:
I will look into why "cl_main.c, CL_ServerExtension_FullUserinfo_f()" doesn't work correctly later today.

Perpixel:
> 
> 
> Don't know what you mean. QSS and vkQuake SV_DropClient do look almost identical. And no, this isn't in QSS so I'm uncertain that it's the right thing to do.
> 
> The advertised protocol needs to stay compatible with other clients, you can't just send extra data.

By different, I mean QSS doesn't use on the //fui message (CL_ServerExtension_FullUserinfo_f) to update player infos. Unless I'm looking at the wrong QSS depot? To me, that's a big difference no?

Perpixel:
> 
> 
> I will look into why "cl_main.c, CL_ServerExtension_FullUserinfo_f()" doesn't work correctly later today.

Ok, I was looking at the wrong QSS depot, my bad.

On drop or connect, the fui command is sent to each client console buffer. I cannot explain the syntax //fui with arguments. Should be "cmd fui" and // are to comment a line in console, no? It was committed by Shpoike on Sep 2, 2020. Incomplete feature?

I stand by my first pull request from the information I gathered.

Novum:
One more thing: Can we get rid of the `// Perpixel - Update scoreboard` comment? I don't like authorship comments. If people really feel strongly about this we should add a section of authors to the readme or something.

Novum:
And so nobody can tell me I'm a hypocrite: 40b6cdca26d966389dd1b7c6204ba8e20b95669f

andrei-drexler:
I know I'm late, but I finally got to take a closer look at the whole `//fui` thing and it seems there are at least 2 reasons why it doesn't work:
1. in vkQuake the string is [passed directly](https://github.com/Novum/vkQuake/blob/22c5d1a484946c3f7758c590a04fef2e6b40a691/Quake/cl_parse.c#L1756-L1758) to the console buffer (which ends up treating it as a comment and ignoring it), whereas QSS [routes it](https://github.com/Shpoike/Quakespasm/blob/8b676fcb73160a4e709cbbbb6ffc7419fe6330cc/quakespasm/Quake/cl_parse.c#L2527-L2529) through a dedicated [CL_ParseStuffText](https://github.com/Shpoike/Quakespasm/blob/8b676fcb73160a4e709cbbbb6ffc7419fe6330cc/quakespasm/Quake/cl_parse.c#L2236-L2283) function that handles `//` differently.
2. even if we were to add that function, in vkQuake `fui` is sent [with an empty string](https://github.com/Novum/vkQuake/blob/22c5d1a484946c3f7758c590a04fef2e6b40a691/Quake/host_cmd.c#L1758) as a second argument in `Host_Spawn_f `, while QSS [passes](https://github.com/Shpoike/Quakespasm/blob/8b676fcb73160a4e709cbbbb6ffc7419fe6330cc/quakespasm/Quake/host_cmd.c#L1863) a [client userinfo](https://github.com/Shpoike/Quakespasm/blob/8b676fcb73160a4e709cbbbb6ffc7419fe6330cc/quakespasm/Quake/server.h#L210) string. vkQuake doesn't have that field, nor the functions that deal with it ([SV_DecodeUserInfo](https://github.com/Shpoike/Quakespasm/blob/8b676fcb73160a4e709cbbbb6ffc7419fe6330cc/quakespasm/Quake/cl_main.c#L1410-L1440), [SV_UpdateInfo](https://github.com/Shpoike/Quakespasm/blob/8b676fcb73160a4e709cbbbb6ffc7419fe6330cc/quakespasm/Quake/cl_main.c#L1441-L1511)).

I think it might be a good idea to remove the `//fui` calls entirely for now (since they're just no-ops that obfuscate the differences versus QSS, hence the comments), along with  `CL_ServerExtension_FullUserinfo_f` / `CL_ServerExtension_UserinfoUpdate_f` / `CL_UserinfoChanged` (effectively unused), and in case there are similar bug reports in the future maybe we should look into adding the whole userinfo machinery from QSS.

For what it's worth, I've also tried the code in this PR in a 4-player match with multiple map changes and can confirm that it works as intended. I've tried a similar match in QSS for comparison, and it looks like the behavior there is the reverse of  #342 : each player only sees the colors of those that were already on the server when they joined.

PR #369, opened 2021-09-19
Added weapon bindings to control options

albrechtjess:
I was able to add the weapon bindings to options with the impulse commands in menu.c. After doing this I noticed the UI doesn't scale to fit the new options. I did some minor adjustments to get them to fit on screen but could still use some improvements as the UI goes off-screen when the scale is turned up too high. I included the original's in comments for quick comparisons.


cmbernard333:
This would be really useful for servers especially in case where you need to start play over in longer map packs.

JosiahJack:
Ideally we'd add support for bindlist.lst and have a default one for vanilla in vkquake.pak.  I'm planning to work on binlist support soon and PR to the various popular engines these days, so that might deprecate this PR.  +1 for the idea.

Novum:
What is `bindlist.lst`?

JosiahJack:
Bindlist is a somewhat obscure feature, original with Quake 2, that lets mods specify the keybinds available in the options menu, letting any custom impulses and such be listed in the menu and able to be set there.  It's been rather infrequently implemented which is somewhat sad.

Example syntax:
```"+attack"         "attack"
"impulse 10"      "change weapon"
"impulse 12"      "prev weapon"
"+jump"           "jump / swim up"
"+forward"        "walk forward"
"+back"           "backpedal"

 ...and so forth.

This lets mods add their own keybinds to the in-game menu like this:
![keep_keybinds](https://user-images.githubusercontent.com/7354604/180890899-d40a497a-a89e-4bcb-9d40-4f594ecfbd3d.png)

Novum:
Fixed by c523480a61a38bb12cb6e24f0750084b71407ff0

PR #374, opened 2021-09-30
ignore more game data

mattyclarkson:
Ignores in tree expansion pack 1/2, arcane dimensions and the unpack of the authentic model collection text file.


Novum:
Uh, what? You aren't supposed to put mods in the source code folder

mattyclarkson:
That's OK, I was being lazy and not installing the binary ☺️

PR #400, opened 2021-11-26
Fix format overflow warning with 32-bit gcc

ahesford:
The move to `-Werror` causes a format overflow warning to kill builds with 32-bit GCC (i686, armv{6,7}l). I have no idea why this doesn't cause a similar problem with 64-bit builds. In any case, make the output buffer large enough to suppress the warning.


sezero:
applied, thanks.

PR #401, opened 2021-11-26
test against all ortho angles, not just v.angle == {0, 0, 0}...

### LINKED ISSUE #375: [Bug] Problem with func_train
It appears vkQuake inherited an [issue from QSS regarding func_train](https://github.com/Shpoike/Quakespasm/issues/75).

Tested with: vkQuake v1.11.0 (x64)

QUINTIX:
... before applying extended collision detection logic.

This fixes #375 

Checking for +/- an arcsecond from right angles is entirely arbitrary means of doing float comparisons in "fixed point" instead comparing directly to `{0,0,0}` (by implicitly converting `float`s to `bool` as was the case). Could just as well be testing against them being within 4096ths of a degree of right angles.

Though it works, as external testing against the specific level in question will prove, I'm not fully confident in this solution.
Should it be transposing components in `absmin` and `absmax` when `IsOrthogonalDeg` is `true`? Should there be more done with entities types that effectively ignore the `origin` value?


Quake/world.c
```diff
@@ -421,7 +421,7 @@ SV_LinkEdict
 void SV_LinkEdict (edict_t *ent, qboolean touch_triggers)
 {
 	areanode_t	*node;
-
+
```
Novum, line 424: Extra whitespace

Novum:
Looks good, thank you. Just maybe revert that extra whitespace.

Quake/world.c
QUINTIX: fix'd

PR #404, opened 2021-12-01
bail on object count != 1

### LINKED ISSUE #402: unable to play (time)demos in 1.12.1 on Mac
After doing a timedemo and switching to another mod it'll give a fps, but I don't see or hear anything, only looking at the console. This problem wasn't there before in previous versions.

`]timedemo demo1
Playing demo from demo1.dem.
]game ad
3095 frames  15.3 seconds 201.7 fps
"game" changed to "id1;ad"`

QUINTIX:
this fixes #402 

Optionally, this could just as easily be `!fread...`, but I prefer to be explicit here.

### DIFF
--- a/Quake/cl_demo.c
+++ b/Quake/cl_demo.c
@@ -114,7 +114,7 @@ static int CL_GetDemoMessage (void)
 	}
 
 // get the next message
-	if (fread (&net_message.cursize, 4, 1, cls.demofile) != 4)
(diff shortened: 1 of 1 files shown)


sezero:
thanks

Novum:
I hate this API. Why are there even two parameters and not just bytes? Anyway, thanks for the fix!

PR #430, opened 2022-02-12
UDP networking: make sure sendNext runs after ack

temx:
Fixes a race where a server hitch of over one second would hang connected clients that were in the middle of receiving a multi-packet reliable message, commonly seen when switching to a large map on a listen server.
The issue is Datagram_ProcessPacket() advances ackSequence and sets sendNext to true, expecting SendMessageNext() to advance sendSequence. But the timeout check happens before that, and ReSendMessage() sets sendNext to false, leaving the connection in an unrecoverable state.


Novum:
I suspect those issues also exist in QuakeSpasm Spiked? Maybe worth reporting there too.

PR #434, opened 2022-02-19
Add the missing pieces for UF_TRAILEFFECT and UF_TAGINFO from QSS

vsonnier:
Those attributes were half-ported probably because pieces get lost when importing QSS FTE extensions. These are the missing pieces making those capabilities working. (in theory)


Novum:
Thanks

Issue #426, opened 2022-02-07
[Bug] A couple of bugs in Arcane Dimensions


TMHKR:
**Bug 1:** Some sounds specific to AD won't play (examples: breakable windows, flying skulls and necromancers in ad_necrokeep)
**Bug 2:** Difficulty level shown on HUD when pressing the Tab key remains "skill 1" regardless of what difficulty you chose

vkQuake: 1.12.2
Arcane Dimensions: 1.80 patch 1
OS: Windows 10 x64

Novum:
One bug per issue please.

TMHKR:
I apologize (new on this website) but at least I hope I made you aware of these bugs.

vsonnier:
> Bug 1: Some sounds specific to AD won't play (examples: breakable windows, flying skulls and necromancers in ad_necrokeep)

Works for me. Load save, break the window, the flying skulls are making noise. Both for 1.12.2 and the current `master`.
[s0.zip](https://github.com/Novum/vkQuake/files/8043649/s0.zip)

Novum:
@vsonnier I feel like I had this happen before to me that certain sounds aren't playing. So I'll keep this open for now, need to investigate at some point.

TMHKR:
**UPDATE:** Bug 1 (sound bug) doesn't seem to occur by using cl_nopext 0 (I had it set to 1)

vsonnier:
> UPDATE: Bug 1 (sound bug) doesn't seem to occur by using cl_nopext 0 (I had it set to 1)

I see that on the `master` too with `cl_nopext 0`. Spiked seems to have the same problem. On the other hand, Ironwail (which has no extensions support at all) works as expected.

vsonnier:
OK, I think the bug is here:

https://github.com/Novum/vkQuake/blob/c8721b2f3bc38d4027b895e2f568873d8a6ef370/Quake/sv_main.c#L1216

This is a much simpler condition that in regular Quakspasm when testing sound capabilities vs. protocols.

Here, as in Spiked, if we have `cl_nopext 1`, then `client->protocol_pext2` is 0 and if we have more than 256 cached sounds (which we have here) this teh sound play is bailed out like if we were in protocol 15.

vsonnier:
Question : 

In `sv_main.c (1216)`, if we change the check by simply:
```c
//PROTOCOL_NETQUAKE do not support more than 256 sounds and/or 8192 entities.
[...]

PR #435, opened 2022-02-19
Fix for missing sounds in protocols > 15

vsonnier:
This is the fix I suggested for #426:
> Bug 1: Some sounds specific to AD won't play (examples: breakable windows, flying skulls and necromancers in ad_necrokeep)

This allows more than 256 sounds independently of protocol extensions presence, except for PROTOCOL_NETQUAKE where it behaves like Quakspasm: such cached sounds > 256 are not played.


Novum:
Thanks

PR #438, opened 2022-02-26
Fix SZ_GetSpace server crashes

### LINKED ISSUE #437: [Bug] ad_mountain "SZ_GetSpace: overflow without allowoverflow set"
When playing Firetop Mountain (ad_mountain) on hard difficulty (skill 2), loading a saved game can fail with an error message.

Steps to reproduce the problem:
1. Play ad_mountain on skill 2 (play normally, don't noclip your way through)
2. Save the game before entering the final lava chamber
3. Start the fight with Chthon and Shamblers
4. Throw yourself into lava and die
5. Attempt to load the saved game

The game might load successfully if you're lucky, but most of the time I encounter the error message:
SZ_GetSpace: overflow without allowoverflow set

I didn't encounter this error on any other maps, though I haven't tested it on Leptis Magna where some people experienced the crash during final battle (Ironwail engine or vanilla QuakeSpasm for example).

vkQuake: 1.12.2 x64
Arcane Dimensions: 1.80 patch 1
System: Windows 10 x64

temx:
This fixes #437. There are two issues: for FTE updates, the code can handle overflows by splitting into multiple packets as long as the buffer has some padding (which it had), but the requested size mistakenly included the padding.
For regular updates, the overflow check assumed 16 bit coords and 8 bit angles, whereas 32 bit coords and 16 bit angles are used by default by protocol 999. Assume the worst case (float coords and angles).

### DIFF
--- a/Quake/sv_main.c
+++ b/Quake/sv_main.c
@@ -1830,7 +1830,9 @@ void SV_WriteEntitiesToClient (client_t *client, sizebuf_t *msg)
 
 		// johnfitz -- max size for protocol 15 is 18 bytes, not 16 as originally
 		// assumed here.  And, for protocol 85 the max size is actually 24 bytes.
-		if (msg->cursize + 24 > maxsize)
+		// For float coords and angles the limit is 36. 
+		// FIXME: Use tighter limit according to protocol flags and send bits.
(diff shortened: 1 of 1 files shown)


Novum:
Thank you

PR #451, opened 2022-04-17
Update .gitignore

sbritorodr:
git will ignore any subfolder inside Quake/. Quake/id1 and any mods (like Quake/honey) will be excluded from this repository.


Novum:
I really dislike this in the first place. Use `-basedir` instead.

PR #455, opened 2022-04-25
Vanilla particle spawn / limit enhancements

temx:
This addresses #452 better than an obscure command line parameter.
Despite the spawn throttling, 16K particles are needed to accommodate 8 gibbed monsters + 2 in-flight rockets.
The CPU cost of 16K active particles is about 0.19ms (Skylake 4GHz)


Novum:
Thanks, this is great.

PR #456, opened 2022-04-26
Mark brush lightmaps for GPU update

temx:
This fixes stale lightning if brush models end up in their own lightmap with no surfaces marking it as modified.


Novum:
Thanks!

PR #392, opened 2021-11-13
Add Mac CI

QUINTIX:
I'm rather new to github actions, having first tried it with [my merge of newer tyrquake into a quake FOV & projection mod called blinky](/QUINTIX/blinky/actions/runs/963009406) by [Shaun Lebron](/shaunlebron)

However, as a followup to https://github.com/Novum/vkQuake/issues/383#issuecomment-955580897
> macOS is not a supported platform because I can't test it.

I've created a CI script for mac in my fork of vkQuake. It successfully builds remotely
https://github.com/QUINTIX/vkQuake/actions/runs/1457190063 (despite not putting a full vulkan SDK in the remote image; not sure how to do that.)

Unlike on my local MacBook Pro with Mac OS Big Sur, where `VULKAN_SDK=~/vulkansdk-macos-1.2.141.2/macOS/`
and where I have ~~who-knows-what version of~~ SDL2 2.0.16 at `/Library/Frameworks/SDL2.framework` and `/usr/local/Cellar/sdl2/`; said local build fails 
```
Undefined symbols for architecture x86_64:
  "_vkGetPhysicalDeviceFeatures2", referenced from:
      _VID_Init in gl_vidsdl.o
  "_vkGetPhysicalDeviceProperties2", referenced from:
      _VID_Init in gl_vidsdl.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [vkquake] Error 1
```
when I run `make`

~~the github-actions generated binary _almost_ runs on my local mac.~~

```
Command line: ./vkquake_ -basedir [redacted]/content
Found SDL version 2.0.16
Detected 4 CPUs.
Built with Clang 13.0.0 (clang-1300.0.29.3)
Quake 1.09 (c) id Software
GLQuake 1.00 (c) id Software
FitzQuake 0.85 (c) John Fitzgibbons
FitzQuake SDL port (c) SleepwalkR, Baker
QuakeSpasm 0.94.2 (c) Ozkan Sezer, Eric Wasylishen & others
QuakeSpasm-Spiked (c) Spike
vkQuake 1.11.1 (c) Axel Gneiting & others
Host_Init
Playing registered version.
Console initialized.
UDP_Init: skipping gethostbyname for [redacted]-MacBook-Pro.local
UDP4 Initialized
UDPv6 Initialized
Server using protocol 999+ (FTE-RMQ)
Exe: 18:32:10 Nov 13 2021
384.0 megabyte heap

[...]

PR #483, opened 2022-06-06
Windows and macOS continuous integration workflows

alexey-lysiuk:


Novum:
There is a problem with Windows CI:
` 'D:/a/vkQuake/vkQuake/VULKAN_SDK\bin\spirv-opt.exe' is not recognized as an internal or external command,
  operable program or batch file.`

alexey-lysiuk:
If it's mandatory, please make it a failure of a build process. When build succeeds without some step(s), I assumed that it's OK to skip it.

QUINTIX:
this makes #392 redundant, and I'm quite happy to see that

PR #492, opened 2022-06-25
Replace remaining custom memory allocation calls

alexey-lysiuk:
These files are not compiled by default, so continuous integration build didn't detect any issues with them


sezero:
Thanks.

Novum:
Cool!

PR #494, opened 2022-06-26
Add meson.build and use on Linux & macos

Novum:


meson.build
```diff
@@ -0,0 +1,171 @@
...
+    'Shaders/Compiled/Release/update_lightmap.comp.c',
+    'Shaders/Compiled/Release/world.frag.c',
+    'Shaders/Compiled/Release/world.vert.c',
+]
+
+cc = meson.get_compiler('c')
+deps = [
+    cc.find_library('m', required : false),
+    cc.find_library('dl', required : false),
+    cc.find_library('pthread', required : false),
```
eli-schwartz, line 111: For dl, you can use (since Meson 0.62.0) https://mesonbuild.com/Dependencies.html#dl-libdl
For pthread, it's always advised to use https://mesonbuild.com/Threads.html

Novum: Thanks, will check. this is my first time using meson

Novum: My system only has Meson 0.61 so this will have to wait. I changed pthread to dependency('threads') though

PR #497, opened 2022-06-28
Enforce warnings as errors in meson build

Novum:


meson.build
```diff
@@ -1,6 +1,6 @@
 project('vkquake', 'c')
 
-cflags = []
+cflags = ['-Wall', '-Wno-trigraphs', '-Wno-unused-function', '-Werror']
```
eli-schwartz, line 3: Meson has a builtin option for this, `meson setup builddir/ --werror --warnlevel 1`. This is recommended as it allows setting the warnings more portably, and because sometimes compiler updates add new warnings so it's valuable to allow people to turn of Werror without actually editing the meson.build.

The recommended way for projects to enforce this as the default is to use the project() function:

```meson
project('vkquake', 'c',
    default_options: ['warning_level=1', 'werror=true'],
)
```

You can use default_options to add any [option](https://mesonbuild.com/Builtin-options.html) (see also `meson configure` to print the list locally). It's very handy. :)

Novum: Thanks, that makes sense

PR #498, opened 2022-06-28
fix MinGW format string 'unknown conversion type character' warnings :

sezero:
```
In file included from mimalloc/static.c:39:0,
                 from mem.c:30:
mimalloc/options.c: In function 'mi_vfprintf_thread':
mimalloc/options.c:349:53: warning: unknown conversion type character 'z' in format [-Wformat=]
     snprintf(tprefix, sizeof(tprefix), "%sthread 0x%zx: ", prefix, _mi_thread_id());
                                                     ^
mimalloc/options.c:349:40: warning: too many arguments for format [-Wformat-extra-args]
     snprintf(tprefix, sizeof(tprefix), "%sthread 0x%zx: ", prefix, _mi_thread_id());
                                        ^~~~~~~~~~~~~~~~~~
```
MinGW builds against msvcrt.dll by default and %z isn't supported there,
as far as I can remember. So, invent and use `MI_PRISZU` and `MI_PRISZX` macros
in mimalloc-internal.h and use them instead.


Novum:
Yes, but also this should be reported as a bug to mimalloc. Will do. Thanks!

PR #514, opened 2022-07-04
replace naive spin wait with native primitive

pg83:
According to "perf record", vkQuake on intro spend 20 - 25% of CPU in https://github.com/Novum/vkQuake/blob/master/Quake/tasks.c#L129

According to https://www.realworldtech.com/forum/?threadid=189711&curpostid=189723 , naive spin wait lock/semaphore/etc should not be used in userspace applications.

After replace, program uses less(approx. 10%) CPU, and I have not notice any latency degradation.

### DIFF
--- a/Quake/tasks.c
+++ b/Quake/tasks.c
@@ -128,27 +128,7 @@ SpinWaitSemaphore
 */
 static inline void SpinWaitSemaphore (SDL_sem *semaphore)
 {
-	int remaining_sleeps = WAIT_SLEEP_COUNT;
-	int remaining_spins = WAIT_SPIN_COUNT;
-	int result = 0;
-	while ((result = SDL_SemTryWait (semaphore)) != 0)
-	{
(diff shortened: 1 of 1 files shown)


Novum:
Intel TBB does the same thing, so do many other task libraries. For a good reason. The timer frequency on Windows is 1ms at it's highest frequency, which is decreasing throughput if you immediately sleep threads on the semaphore.

pg83:
Maybe I should apply this logic only for __linux__ case?

Novum:
We can tweak the spin/sleep count to a more sensible level. Might still be too high.

pg83:
>We can tweak the spin/sleep count to a more sensible level.

Sure, It will help, to some extent. May be, 100 * 3 == 300 iteration will suffice for windows.

Novum:
I removed the sleeps and made it 100 spins

Issue #517, opened 2022-07-06
[bug?] In palettised mode the colours should be applied per texel


theptoing:
In the original software Quake versions the lightmap never bisects a texture pixel, meaning any given texel can only be 1 of the 256 indices. This actually makes the lighting look a lot more integrated with the textures (and imo it would be nice to have a feature that allowed to toggle this on and off even in 24 bit lighting mode, namely because of the added coherency of the resulting texture+light)

Novum:
It's way more complicated to implement, but technically this is true.

temx:
Since textures and lightmaps are aligned (and lightmaps are 1/16 of the texture size), the quick and dirty way to do this is just quantize the lightmap coordinates in the world frag shader:

```
	ivec2 lm_size = textureSize(lightmap_tex, 0).xy;
	vec2 uv_exp = floor(lm_size * in_texcoords.zw * 16) / 16 / lm_size;
	vec3 light = texture(lightmap_tex, uv_exp).rgb * 2.0f;
```

Novum:
2e29b2a01db9aae0bd74b02c378f9cd43d575f11 mostly addressed this, thanks @temx 

Fog will still cause it to go wrong, but I think this is acceptable.

PR #531, opened 2022-07-16
Per-texel lightmap for 8-bit mode

temx:
Possible solution for #517

Sample [[scene]](https://user-images.githubusercontent.com/24851409/179371741-1bdc87b3-e9ec-45eb-aef0-ede3289b6a26.png) [[before]](https://user-images.githubusercontent.com/24851409/179371742-c807a1dd-a75d-4953-a516-415f7356b675.png) [[after]](https://user-images.githubusercontent.com/24851409/179371746-044ef6ca-452d-4b55-913c-cc238671527a.png) (this is without palettization for clarity, but it only triggers with `vid_filter 1` `&&` `vid_palettize 1`)


Novum:
We probably want to quantize fog as well? Otherwise this will cause the same issue?

Novum:
Okay I understand why we are not dealing with fog. It's way  more complicated and probably needs tangent space etc. Not going there. This is helping already, so it's a good addition.

temx:
I tried a screen space derivatives hack for fog but it got unwieldy indeed. 
We can hope the fog won't be too dense, and in the distance it doesn't matter much.

PR #533, opened 2022-07-20
MacOS fixes

robert-house:
Here are the changes to get MacOS builds working again. When testing on Windows and Linux, I discovered that the portability extensions haven't been enabled yet, so in the mean time, I have introduced an APPLE define. There appear to be some breaking changes coming down the pipe for Vulkan with regards to api on api wrappers (MoltenVK, DXVK, etc.), so this would have needed to be done anyways.

On line 1058, I had to set the extension name as a string constant as it appears the enum const VK_KHR_PORTABILITY_SUBSET_EXTENSION_NAME is NOT available in this version of the SDK. In a future release, I will change this assuming they add it to the non-beta header.

I didn't include the recompiled *.c shader files. Mac users will have to run the compile script to fix the shader issue.

### DIFF
--- a/Quake/gl_rmisc.c
+++ b/Quake/gl_rmisc.c
@@ -691,9 +691,9 @@ static void R_InitDynamicVertexBuffers ()
 	vkGetBufferMemoryRequirements (vulkan_globals.device, dyn_vertex_buffers[0].buffer, &memory_requirements);
 
 	const int align_mod = memory_requirements.size % memory_requirements.alignment;
-	const int aligned_size = ((memory_requirements.size % memory_requirements.alignment) == 0)
-								 ? memory_requirements.size
-								 : (memory_requirements.size + memory_requirements.alignment - align_mod);
+    const int aligned_size = ((memory_requirements.size % memory_requirements.alignment) == 0)
+                             ? memory_requirements.size
+                             : (memory_requirements.size + memory_requirements.alignment - align_mod);
 
 	VkMemoryAllocateInfo memory_allocate_info;
 	memset (&memory_allocate_info, 0, sizeof (memory_allocate_info));
@@ -757,8 +757,8 @@ static void R_InitDynamicIndexBuffers ()
 
 	const int align_mod = memory_requirements.size % memory_requirements.alignment;
 	const int aligned_size = ((memory_requirements.size % memory_requirements.alignment) == 0)
-								 ? memory_requirements.size
-								 : (memory_requirements.size + memory_requirements.alignment - align_mod);
+                                 ? memory_requirements.size
+                                 : (memory_requirements.size + memory_requirements.alignment - align_mod);
 
 	VkMemoryAllocateInfo memory_allocate_info;
 	memset (&memory_allocate_info, 0, sizeof (memory_allocate_info));
@@ -822,8 +822,8 @@ static void R_InitDynamicUniformBuffers ()
 
 	const int align_mod = memory_requirements.size % memory_requirements.alignment;
 	const int aligned_size = ((memory_requirements.size % memory_requirements.alignment) == 0)
-								 ? memory_requirements.size
-								 : (memory_requirements.size + memory_requirements.alignment - align_mod);
+                                 ? memory_requirements.size
+                                 : (memory_requirements.size + memory_requirements.alignment - align_mod);
 
 	VkMemoryAllocateInfo memory_allocate_info;
 	memset (&memory_allocate_info, 0, sizeof (memory_allocate_info));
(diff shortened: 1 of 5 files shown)


sezero:
Whitespace changes aren't nice and make the patch unreadable

>  I have introduced an APPLE define

Why is the standart `__APPLE__` not enough??

Linkage to libvulkan is available and MoltenVK is not needed anymore?

Quake/gl_rmisc.c
```diff
@@ -692,8 +692,8 @@ static void R_InitDynamicVertexBuffers ()
 
 	const int align_mod = memory_requirements.size % memory_requirements.alignment;
 	const int aligned_size = ((memory_requirements.size % memory_requirements.alignment) == 0)
-	                             ? memory_requirements.size
-	                             : (memory_requirements.size + memory_requirements.alignment - align_mod);
+								 ? memory_requirements.size
+								 : (memory_requirements.size + memory_requirements.alignment - align_mod);
```
sezero, line 696: Please lose whitespace changes

Quake/gl_rmisc.c
```diff
@@ -822,8 +822,8 @@ static void R_InitDynamicUniformBuffers ()
...
```
sezero, line 826: Please lose whitespace changes

Quake/gl_vidsdl.c
```diff
@@ -105,7 +105,7 @@ cvar_t                          vid_fsaamode = {"vid_fsaamode", "0", CVAR_ARCHIV
 cvar_t                          vid_gamma = {"gamma", "0.9", CVAR_ARCHIVE};       // johnfitz -- moved here from view.c
 cvar_t                          vid_contrast = {"contrast", "1.4", CVAR_ARCHIVE}; // QuakeSpasm, MarkV
 cvar_t                          r_usesops = {"r_usesops", "1", CVAR_ARCHIVE};     // johnfitz
-                                                                                  // Vulkan
+																				  // Vulkan
```
sezero, line 108: Please lose whitespace changes

robert-house:
Sorry about that, I didn't catch the whitespace changes.

To answer your question from before: Yes, linking against vulkan will automatically load MoltenVK. Otherwise, the validation layers will not work.

robert-house:
OK thanks for taking a look. Sorry about the formatting issues! Going forward, I'll be more careful (and make sure CLion diffs whitespace and stops autoformatting as well)

Novum:
Why exactly is the portability subset extension even needed? Nothing in the patch makes that clear to me.

robert-house:
In short: as of sdk version 1.3.216.0, Vulkan (using MoltenVK) will fail to initialize an instance without VK_KHR_PORTABILITY_ENUMERATION_EXTENSION. Here is a link to the release notes saying such as much: https://vulkan.lunarg.com/doc/sdk/1.3.216.0/mac/release_notes.html. As for the VK_KHR_PORTABILITY_SUBSET extension, I was getting a validation error without enabling in on device creation. Here is what I saw:

Validation Error:  vkCreateDevice: VK_KHR_portability_subset must be enabled because physical device VkPhysicalDevice 0x60000379fe80[] supports it The Vulkan spec states: If the VK_KHR_portability_subset extension is included in pProperties of vkEnumerateDeviceExtensionProperties, ppEnabledExtensionNames must include "VK_KHR_portability_subset"

I enabled it and it went away. Thats about as far into the matter I looked as I was mainly trying to figure out the descriptor problem. Speaking of which, I am kind of confused as to why the previous limits were not causing the same issue on other platforms. My guess is the drivers are doing something different behind the scenes, but I only have a Nvidia card to test with and it was working fine before I made the changes (and after!). Allocation of the screen_effects set should have failed  with VK_KHR_OUT_OF_POOL_MEMORY on Windows and Linux as well. My understanding could be completely off, but this is what made sense to me.

I wish I had a better answer for you regarding the shaders. Unfortunately, I don't have too much insight into WHY cross compiling with binaries provided on other platforms is failing and I lack the required skillset to get to the root of the problem. It has to be some bug with SPIRV-Cross. The output when I ran the shaders through the SDK provided glslangValidator and then bintoc were vastly different in size compared to the versions that are provided by default. The zsh script is more of a bandaid to at least get it working in the mean time until a root cause can be determined.

Novum:
The runtime can choose to ignore pool limits completely, and in fact NVidia does ignore it. It is valid behaviour.

Novum:
I believe I know what's going on. At some point I added build steps that run `spirv-opt` and `spriv-remap` on the release shaders to reduce their size. This likely triggers a bug in SPIRV-cross which makes MoltenVK fail. Need to check if I can find a workaround.

Novum:
After looking into the portability subset it seems like you ran into this issue because you use the Vulkan loader. I don't see the point of using the loader when the only Vulkan device that will ever exist on macos is MoltenVK?

Novum:
@robert-house Thanks for testing though! I appreciate it a lot.

PR #536, opened 2022-07-24
Add wave codec support to meson

### LINKED ISSUE #535: Meson missing wave codec support
When building with meson, no wave codec is built.

This can cause music to not play for players who are using wav music files instead of ogg/flac.

daftmugi:
Fixes #535.


sezero:
Thanks.

PR #539, opened 2022-07-27
Add Model Scale Support

JosiahJack:
Add support for models to use `scale` field. Used to scale up or down misc_model decorations such as statues or trees to add variety and emphasis.

This is purely scaling the model drawing. Scales the bounds to prevent the entity becoming invisible when not looking at its origin.

**Test Files**
[scaletest.zip](https://github.com/Novum/vkQuake/files/9245746/scaletest.zip)

Tiny test mod containing a test map maps/test_scale.bsp and a v1.06 progs.dat with only `.float scale;` added to defs.qc
Disregard the fact that this is an enemy model; I'm merely using a func_illusionary to display it and chose a standard id1 model.

Expected results below.  The doubled one in the middle is scale 1.0 (not set).  The unscaled one at the far left is scale set too small.
image

### DIFF
--- a/Quake/cl_parse.c
+++ b/Quake/cl_parse.c
@@ -1203,6 +1203,8 @@ static void CL_ParseUpdate (int bits)
 			ent->alpha = ent->baseline.alpha;
 		if (bits & U_SCALE)
 			ent->netstate.scale = MSG_ReadByte (); // PROTOCOL_RMQ
+		else
+			ent->netstate.scale = ent->baseline.scale;
 		if (bits & U_FRAME2)
 			ent->frame = (ent->frame & 0x00FF) | (MSG_ReadByte () << 8);
 		if (bits & U_MODEL2)
@@ -1242,7 +1244,10 @@ static void CL_ParseUpdate (int bits)
 			ent->alpha = ent->baseline.alpha;
 	}
 	else
+	{
 		ent->alpha = ent->baseline.alpha;
+		ent->netstate.scale = ent->baseline.scale;
+	}
 	// johnfitz
 
 	// johnfitz -- moved here from above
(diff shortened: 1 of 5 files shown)


temx:
`sv_main.c`: the changes seem to be redundant with lines 794-797 (just above). The `39` in line 1899 should be raised to `40`.

Novum:
Please squash this

JosiahJack:
> Please squash this and run clang-format.

Done.  If you don't mind my asking, why squash it now when it can be squashed at merge?

sezero:
>  If you don't mind my asking, why squash it now when it can be squashed at merge?

Because otherwise it's rude, and severely hinders other ways of testing PR patches
E.g. downloading the patch `https://github.com/Novum/vkQuake/pull/539.patch` and
attempting to apply it to your local tree would fail.

Not everything is about github gui.

Novum:
It's also easier to keep track of what was actually changed

sezero:
Josiah submitted a similar patch to qs too, but I don't have anything to test it with so I requested reviews from @ericwa or @andrei-drexler

JosiahJack:
> ...but I don't have anything to test it with...

Ah, good point.  Added test files and expected results to the top post.

sezero:
Haven't tested this vkquake version of the patch, but the test map/progs you gave above rendered broken for qs -- see https://github.com/sezero/quakespasm/pull/32#issuecomment-1202006404

JosiahJack:
> ...the test map/progs you gave above rendered broken...

I accidentally included the wrong progs.dat by mistake, in addition to missing a discrepancy over in standard QS.  The zip has been updated in the top post to use the progs.dat I've been using during testing.  Sorry for the hassle.

JosiahJack:
Add scaling to R_AddEfrags based on comment [here.](https://github.com/andrei-drexler/ironwail/pull/110#issuecomment-1198938902)

Also fixed r_showtris 1 to draw with scale applied.

Novum:
Closing this. Will port if it eventually lands in QuakeSpasm upstream.

PR #569, opened 2022-10-06
Replace atomics with thread-local variables

temx:
On a 6c/12t Skylake this appears to be quite a bit faster in some cases (1.23 -> 1.01ms in alk_dancing start, 12.12 -> 10.81ms in shib8 start, no change in ad_tears).


Novum:
Fixed by 06e9fb789a2efd57ca196a293f84f18e7a0ef19b with simpler solution

PR #572, opened 2022-10-30
Cubemap sky support

temx:
@Novum this adds cubemap texture sky support for [seamless filtering across faces](https://user-images.githubusercontent.com/24851409/198900321-838f6961-42ee-4099-aa5c-426e95ff1ca3.png) (currently it just [chops half a texel](https://user-images.githubusercontent.com/24851409/198900323-d8ff7182-7155-422d-8a74-da59ccb6d2d1.png)) and also draws the sky brushes directly with no stencil pass (about 5% faster on Vega and 4% on Gen9.5 and Turing in `alk_dancing`).


Novum:
The stencil was necessary in my tests otherwise there are edge cases where things break.

temx:
Was this using a cubemap texture? I would expect it'd be tricky otherwise. At any rate I don't see any problems in ad_tears.

Novum:
So the non-stencil path only applies to the cubemap? Sorry I didn't look that closely at the code yet.

Novum:
I see, there is a `qboolean Sky_NeedStencil ()`

Novum:
Before merging this, why is this a thing in the first place? Has someone in the community asked for it?

temx:
It's not a new feature, it's a better way to draw the sky that's already used in existing levels (a bit more efficient, and better filtered). But it doesn't expose any new functionality.

Novum:
And this is for quality? Or what's the difference? Performance of drawing 12 tris instead should be the same? Stencil could be skipped either way I believe (with enough depth bias perhaps)?

temx:
For eligible skies (6 square faces of the same size) differences are:
* sky brush tris are drawn with the final color directly (before: sky brush tris are drawn to stencil with no color, then the sky is drawn as a stencil-tested large enclosing box)
* box edges & corners get cubemap filtering (before: half a texel is removed from each side to make the seam less visible, but it's still there if you know where to look)

The extra cost of the old method is at least a stencil clear, write (for what's covered by the sky) and test (for the area covered by the sky bounds - often the entire viewport).

This also fixes issues with the corners of the enclosing box getting clipped with very large values of gl_farclip (there's no longer an enclosing box).

Novum:
Okay, sounds good

PR #573, opened 2022-11-03
Allow exceeding TEXTURE_HEAP_SIZE if necessary

temx:
@novum It turns out there are maps using 6x2Kx2K skies that need 96MB+ by themselves. It's possible to fall back to the old path or downsize them in that case, but this removes the size limitation altogether by allocating a big enough heap for images that wouldn't fit otherwise. Does this seem OK to you?


Novum:
Yeah, makes sense. Although people really shouldn't be using 96MB textures in a Quake mod.

PR #579, opened 2022-11-25
Indirect brushes

temx:
@Novum This implements GPU compute index buffer generation for most brush surfaces, which decreases CPU cost significantly with the corresponding speedup if CPU bound.

On extreme scenes the reduction can be >80%, making it possible to draw 500K surfs using ~3ms of total (single-threaded) CPU time.

All world polys are supported including skies and water. Leafs are frustrum culled in the CPU and the resulting surface bitmap is put into a double buffered host-visible buffer. The GPU does backface culling, increases the indirect draws' `indexCount` and places the indices in a preallocated worst-case index buffer (18 MB for `shib8`, 5 MB for `ad_tears`). This is the only significant new GPU allocation, since the per-surf call and index data is packed into the existing `lm_compute_surface_data_t` struct. For brush models, non-transformed opaque world submodels are supported. These are painted in the surface visibility bitmap, anything else is rendered as before.


Novum:
Nice! 

Some background: I gave this a shot once before, but I tried to merge together stuff into one gigantic index buffers similar to what we did at id. It worked, but the problem was that the draw order actually needs to be stable for Quake because of transparent brushes.

I assume this is more like Ironwails approach?

PR #592, opened 2022-12-11
Fullscreen mouse cursor

temx:


Novum:
You are correct, this works fine.

PR #595, opened 2022-12-12
Menu tweaks

temx:
- Add FOV slider
- Mouse support for sliders
- Mouse support for scrollbars

### DIFF
--- a/Quake/menu.c
+++ b/Quake/menu.c
@@ -109,7 +109,23 @@ static int scrollbar_size;
 
 void M_ConfigureNetSubsystem (void);
 
-extern cvar_t scr_fov;
+extern qboolean keydown[256];
+extern cvar_t	scr_fov;
+
+static qboolean slider_grab;
+
+/*
+================
+M_GetScale
+================
+*/
+float M_GetScale ()
+{
+	static float latched_menuscale;
+	if (!slider_grab)
+		latched_menuscale = scr_menuscale.value;
+	return latched_menuscale;
+}
 
 /*
 ================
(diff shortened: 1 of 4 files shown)


Quake/menu.c
```diff
@@ -1157,20 +1185,37 @@ void M_Menu_Options_f (void)
...
+		else if (f > 130)
+			f = 130;
+		Cvar_SetValue ("fov", f);
+		break;
 	case OPT_SCALE: // console and menu scale
 		if (scr_relativescale.value)
 		{
 			l = 3.0f;
-			f = scr_relativescale.value + dir * .1;
+			if (mouse)
```
Novum, line 1215: I think we should add a function for those to remove the repetition of this if/else pattern?

temx: Yeah, it should be better. I've factored it out now.

Novum:
Looks good, thanks. Still want to verify at home before merging.

Novum:
So I tested this and what I didn't like is that apparently one can't drag the sliders? Kind of unexpected.

temx:
I added dragging support, and also an option for autoload (and changed its default). If you don't want it it can be omitted.

Novum:
One more request, I think the sliders should snap to the values accessible via keyboard?

temx:
Screen size will now snap above 100. It's a bit strange but no reason not to be continuous in the < 100 range. The others are continuous things so I left them as-is (you could already set any value you wanted in the console). If you think any sliders should be discrete it's a matter of changing the corresponding 999 value to 0. Also keyboard adjustments always re-snap now.

Novum:
Thanks, feels great now.

Issue #548, opened 2022-08-09
[Feature request] Double click demo file to play it


timbergeron:

temx:
Assuming this is from the OS shell, you can set your `.dem` handler to launch vkQuake with the appropriate parameters. For example in Windows you can use a `cmd` file like
```
for %%F in ("%~1\..") do set modname="%%~nxF"
cd /D "%~dp0"
vkQuake -game %modname% +playdemo "%~nx1"
```
place it along vkQuake and assign `.dem` files to it (or just drop them on top of it).

I do like the feature some Doom source ports have where they do the right thing when you directly pass a file as a parameter. Quake could interpret directories as mods, `map` files as `game` + `map`, and `dem` files as `game` + `playdemo` requests.

timbergeron:
added this for dragging a demo into windowed mode:

https://github.com/timbergeron/QSS-M/commit/a1ee1d29e5c44ff9f8f25c22b6c8d6c2f0e6c9c6

timbergeron:
Found this:
https://github.com/timbergeron/Engine-X/blob/1f368b93ccc2ef2a489f1767f26fad977851479c/Engine_X550_source/sys_win_interact.c#L434

vsonnier:
Not interested in adding this.

PR #596, opened 2022-12-12
Drag and drop support for maps, mods, savegames and demos

temx:
This is a possible fix for #548. It only supports stuff from the normal search path.

### DIFF
--- a/Quake/common.c
+++ b/Quake/common.c
@@ -1231,8 +1231,8 @@ being registered.
 */
 static void COM_CheckRegistered (void)
 {
-	int			   h;
-	int			   i;
+	int h;
+	int i;
 
 	COM_OpenFile ("gfx/pop.lmp", &h, NULL);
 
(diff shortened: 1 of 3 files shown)


Novum:
I'm kind of on the fence about this one. Kind of odd to do drag and drop and then you can't actually drag in any file?

Maybe we should implement some menus? Just not sure where they would go.

temx:
Yeah, I'm not so sure either.

I also [considered adding this for commandline parameters](https://github.com/temx/vkQuake/commit/92c18437e0d9ad891573be2a92732c11462af6af) but that's even worse: the application starts in the dragged file's directory, so if you want to support "drag into the exe" then you need to check for that and change dirs.

Supporting external loose files is somewhat awkward - I guess you can try to load them with the currently running mod and hope for the best. Maps come in `.bsp`+`.lit`(+`.ent`) pairs, so that also needs handling.

For menus there's the added problem of having to match the font / bailing out for any mods that modify it. JoeQuake for example [puts everything in the main menu](https://user-images.githubusercontent.com/24851409/207988426-f26194ef-560c-436e-bc1c-c575fc4287fd.png) and [forces it to keep the original style](https://user-images.githubusercontent.com/24851409/207988471-5efbd255-3bfe-43c8-8c11-71724d6f33c5.png). Perhaps an inconspicuous place would be the corner of the screen?

PR #598, opened 2022-12-17
Add Speedometer

Jbudone:
- Introduced speedometer similar to JoeQuake, useful for speedrunners
- Updated character draws to allow size specification
- Added helpful calls -- find best matching color in palette; find hud size for offsetting

### DIFF
--- a/Quake/gl_screen.c
+++ b/Quake/gl_screen.c
@@ -726,7 +726,7 @@ void SCR_DrawSpeed (cb_context_t *cbx)
 	char st[4];
 	float speed = VectorLength (cl.velocity);
 	sprintf (st, "%-3d", (int)speed);
-	int speedWidth = (int)(MIN(1.f, speed / maxSpeedFillWidth) * width);
+	int speedWidth = (int)(fmin(1.f, speed / maxSpeedFillWidth) * width);
 
 	static int bgColor = -1;
 	static int fillColor = -1;
--- a/Quake/gl_screen.c
+++ b/Quake/gl_screen.c
@@ -89,6 +89,7 @@ cvar_t scr_conscale = {"scr_conscale", "1", CVAR_ARCHIVE};
 cvar_t scr_crosshairscale = {"scr_crosshairscale", "1", CVAR_ARCHIVE};
 cvar_t scr_showfps = {"scr_showfps", "0", CVAR_NONE};
 cvar_t scr_clock = {"scr_clock", "0", CVAR_NONE};
+cvar_t scr_showspeed = { "scr_showspeed", "1", CVAR_NONE};
 // johnfitz
 cvar_t scr_usekfont = {"scr_usekfont", "0", CVAR_NONE}; // 2021 re-release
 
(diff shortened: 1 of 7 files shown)


Quake/gl_texmgr.c
```diff
@@ -566,6 +566,33 @@ void TexMgr_LoadPalette (void)
 	((byte *)&d_8to24table_conchars[0])[3] = 0;
 }
 
+/*
+================
+TexMgr_NearestColor
```
Novum, line 571: This doesn't belong here. Has nothing to do with texture management.

Jbudone: Agreed -- what would be a better spot for this? gl_draw.c maybe?

Jbudone: Or maybe common.c?

temx: Are there plans to allow for changing colors? This seems to be only ever called with the pair of values (20, 20, 0), (40, 30, 15), so you could assume the default Quake palette is used and do `bgColor = 51; fillColor = 19;`. Otherwise  `gl_draw.c` may be the best place.

temx:
A few general comments:
* This should be not enabled by default, especially if the cvar is not archived.
* I haven't looked into this closely, but it seems to be doing its own positioning and text scaling. Would it be possible to use a canvas that already has the correct scaling? That way `Draw_FillCharacterQuad_WithSize` wouldn't be necessary, and maybe the positioning could be simplified too.
* The vertical component of the speed is taken into account. I thought the standard was showing only the horizontal components (for example, JoeQuake shows a speed of 0 if you're jumping in-place).

Jbudone:
The abstractions (_WithSize, _NearestColor) were intentional :) its only called here yes, but helpful for future newcomers like myself to not have to hardcode (eg. looking up palette and deciding color, or not being able to scale text)

andrey-budko:
I like the idea. scr_showspeed is the only reason I use my own fork. My implementation is simpler - just print the speed under crosshair.

https://github.com/andrey-budko/vkQuake/commit/45245cf4a904735aed00f5ea777a0e7a9b1979b6

<details>

  image
</details>

Jbudone:
Poke @Novum  -- Is there anything else that needs to change?

- I'd prefer to keep abstractions (_WithSize, _NearestColor) which feel like improvements to the engine
- I can move TexMgr_NearestColor ; but where does that fit better? common.c maybe?
- temx pointed out that I had it on by default (oops!) which I will fix in follow-up

Quake/gl_screen.c
```diff
@@ -698,6 +700,47 @@ void SCR_DrawClock (cb_context_t *cbx)
...
+	float textHeight = 8.f * scale;
+	float textY = y + (height / 2) - (textHeight / 2);
+	float textX = x + textLeftOffset;
+
+	char  st[4];
+	float speed = VectorLength (cl.velocity);
+	sprintf (st, "%-3d", (int)speed);
+	int speedWidth = (int)(fmin (1.f, speed / maxSpeedFillWidth) * width);
+
+	static int bgColor = -1;
```
Novum, line 730: Why are those static?

Quake/gl_screen.c
```diff
@@ -698,6 +700,47 @@ void SCR_DrawClock (cb_context_t *cbx)
...
+	static int fillColor = -1;
+	if (bgColor == -1)
+	{
+		bgColor = TexMgr_NearestColor (20, 20, 0);
```
Novum, line 734: What's the point of this? Can't you just hard code the closest color from the palette? The palette is fixed as far as I know? Why does it even have to be a palette color?

PR #601, opened 2022-12-19
Wip keymodifiers

Jbudone:
NOTE: I'm not looking to submit this yet -- still WIP

I'm looking for some early feedback to see if this is something that would get accepted and merged @Novum

### DIFF
--- a/Quake/draw.h
+++ b/Quake/draw.h
@@ -39,6 +39,7 @@ void	Draw_Fill (cb_context_t *cbx, int x, int y, int w, int h, int c, float alph
 void	Draw_FadeScreen (cb_context_t *cbx);
 void	Draw_String (cb_context_t *cbx, int x, int y, const char *str);
 void	Draw_String_3D (cb_context_t *cbx, vec3_t coords, float size, const char *str);
+void    Draw_String_WithSize (cb_context_t *cbx, int x, int y, const char *str, float size);
 qpic_t *Draw_PicFromWad2 (const char *name, unsigned int texflags);
 qpic_t *Draw_PicFromWad (const char *name);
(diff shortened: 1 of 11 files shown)


temx:
What is the intended purpose and how does this compare to using alias (`alias +shiftdown "bind F2 command_f2_down;bind F3 command_f3_down"; alias -shiftdown "bind F2 command_f2_up;bind F3 command_f3_up"; bind shift "+shiftdown"`)?

Jbudone:
Oh interesting! I didn't know of this trick

The purpose is to allow binding with shift (and potentially more). It feels silly this day in age to have an engine that doesn't natively support keybind modifiers
```
; save/practice spot 1
bind "SHIFT+F1" "save 1"
bind "F1" "load 1"

; save/practice spot 2
bind "SHIFT+F2" "save 2"
bind "F2" "load 2"
```

If aliasing is already an option then this is merely a less hacky solution, cleaner solution

Jbudone:
Poke @Novum  I left the CI issues until I can get some feedback. I'm interested to hear if this is something we'd like for the engine, or if I'm just wasting time here :smile: 

I personally think its a fairly simple change, low to no maintenance, and feels like a normal expectation to allow modifiers for controls. @temx mentioned a hacky alternative that works, so this doesn't solve anything new, but others like me may not realize that trick exists, and it feels silly to depend on a hacky trick when we could support it directly in the engine

Novum:
What's the actual use for this? I haven't seen many games support modifiers. It also adds quite a bit of possible new bug surface.

Jbudone:
temx asked the same thing above (my answer in previous comment). Its convenience without having to depend on the trick he mentioned

PR #606, opened 2022-12-28
[CI] Put platform in built artifact name to prevent overwriting

Macil:
Github Action artifacts are namespaced by workflow, so workflows that have multiple jobs that each upload an artifact with the same name will trample over each other. The build-mingw and build-windows workflows use multiple jobs for creating 32 bit and 64 bit executables, but then name the artifacts the same, causing only one set of artifacts to be retrievable per workflow. This PR makes it so each job gives a descriptive name to its own artifacts.


Novum:
Makes sense

PR #607, opened 2022-12-28
Meson build: Support ARM Macs by default

Macil:
This PR makes it so that the Mac build instructions of `meson build && ninja -C build` just work even on modern ARM Macs where Homebrew installs libraries to `/opt/homebrew/lib`. This makes the extra instructions in the readme for ARM Macs unnecessary. The build still works for non-ARM Macs too.

(Only the `cc.find_library()` call for MoltenVK had to be changed. It was the only `cc.find_library()` call for a brew-installed dependency. The other brew-installed dependencies were found through the `dependency()` function, which I think is able to find brew-installed libraries by default.)

This is basically the same change I made to the [Makefile](https://github.com/Novum/vkQuake/pull/330/files#diff-9b00d1a32e360c0662d3221add6af44af67bfa593236b39c8926da6f522e2f41R155) in https://github.com/Novum/vkQuake/pull/330 redone for the new Meson build script.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -148,7 +148,7 @@ deps = [
 ]
 
 if build_machine.system() == 'darwin'
-    deps += cc.find_library('MoltenVK', required : true)
+    deps += cc.find_library('MoltenVK', required : true, dirs : ['/opt/homebrew/lib'])
 else
     deps += dependency('vulkan')
 endif
--- a/meson.build
+++ b/meson.build
@@ -148,7 +148,11 @@ deps = [
 ]
 
 if build_machine.system() == 'darwin'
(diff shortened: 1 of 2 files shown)


Macil:
Huh, CI is telling me that I'm wrong about this not breaking non-ARM Macs. I'll figure this out and update the PR soon.

Macil:
It turns out that Meson's `find_library()` function [actually skips checking normal library locations if you specify any "extra" directories to check](https://github.com/mesonbuild/meson/blob/1.0.0/mesonbuild/compilers/mixins/clike.py#L1126), so it's not actually safe to list extra directories to check that aren't always used. I've fixed my PR for this now.

PR #612, opened 2023-01-08
Shadows: clearer menu description and persist value

temx:


Novum:
I'm not quite happy with the naming yet, but I'm not sure if this proposal is better

Novum:
Problem is always the limited amount of characters available in the menu.

PR #630, opened 2023-02-25
Fix QuakeC multicasting extension

ilikebreadtoomuch:
vkQuake, trough checkextension, reports to the QuakeC that it supports `FTE_QC_MULTICAST`, yet trying to write a message to MSG_MULTICAST throws an error. Looks like it has the Quakespasm-spiked code for sending multicast messages but WriteDest in pr_cmds.c was never edited to allow MSG_MULTICAST write calls. The new switch cases are directly copied from QSS


temx:
Applied after `clang-format`, thanks.

Issue #631, opened 2023-02-26
Rendering bugs in Dwell 2


penasea:
**Describe the bug**
The transparent water in Dwell 2 d2m6.bsp shows rendering bugs when moving around:
[Video 1](https://user-images.githubusercontent.com/126418320/221432720-8417b8cf-7c45-4862-b0fb-9374561a09e8.webm)
And some of the stones always look odd:
[Video 2](https://user-images.githubusercontent.com/126418320/221432744-d377f551-2de9-4da6-a4ce-a3058a83bd91.webm)

**To Reproduce**
Load Dwell 2 (game dwellv2p2), load the map (map d2m6), look around.

**Expected behavior**
Correct rendering.

**Screenshots**
See above.

**Desktop**
 - OS: Windows 10
 - Nvidia GeForce 1060 6 GB
 - Driver version 528.49

Novum:
What version is this? There has been a bug fix already that sounds like it's the same problem: https://github.com/Novum/vkQuake/commit/4c89ff1059906f36e51fd9a3929814776b18dbb4

temx:
4c89ff1 shouldn't affect this. The flickering is due to entities being added in the order of their first visible leaf; if they span more than one leaf the order depends on visibility (incl. frustum culling). The other stuff is just alpha drawn in the wrong order.

Novum:
Why does this not happen in Quakespasm then? Or does it?

temx:
It does affect QuakeSpasm, I think it renders exactly the same there.
IW has three modes (OIB, distance sorted, and "QS-like" but stable). The rerelease also uses OIB.

Novum:
ef8260517e68e4f523c0305aa482677aa53849f8

PR #632, opened 2023-02-27
Best-effort alpha sorting

temx:
Potential fix for #631

This implements transparency sorting, using 3 methods:

* Alpha entities are classified as underwater or overwater according to their center, and drawn before or after the water (this also helps many other levels when underwater, as it avoids transparents over the water surface getting drawn after the water)
* Alpha entities are sorted back to front by their closest point (view angle independent)
* Transparent brush model surfs are chained in depth order using BSP traversal. This gives perfect ordering within a model for cheap (as long as it uses a single texture as implemented, since it doesn't flush on texture changes). D2M6 needs this because several crystals and waterfalls are grouped as a single entity.

This is not exhaustive but solves all sorting problems I can see in D2M6 (except a few subtle ones caused by the opaque water used as a ceiling - this is fixable but I'm not sure it's worth it) and a few other levels.

### DIFF
--- a/Quake/cl_main.c
+++ b/Quake/cl_main.c
@@ -57,8 +57,11 @@ lightstyle_t	cl_lightstyle[MAX_LIGHTSTYLES];
 dlight_t		cl_dlights[MAX_DLIGHTS];
 
 int		   cl_numvisedicts;
+int		   cl_numvisedicts_alpha_overwater;
+int		   cl_numvisedicts_alpha_underwater;
 int		   cl_maxvisedicts;
 entity_t **cl_visedicts;
+entity_t **cl_visedicts_alpha;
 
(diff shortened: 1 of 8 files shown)


Novum:
What's the performance impact? Thanks for implementing this

temx:
I get around ~1% perf loss in `ad_tears` (~6 microseconds of frame time), 0.5% in `ad_sepulcher`, nothing in `d2m6` itself, and 8% in [`aesop_power`](https://www.quaddicted.com/reviews/aesop_power1.1.html) (0.2 milliseconds). This last one has over a thousand visible alpha brush entities (this is very atypical). The sorting task itself takes ~40 microseconds, most of the extra time appears to be spent traversing the BSP trees.

Novum:
Okay, that's acceptable.

PR #646, opened 2023-03-12
Simplify Always Run menu option to on|off

Macil:
This changes the "Always run" options from off/vanilla/quakespasm to just on/off, because the differences between vanilla and quakespasm are insignificant, mostly irrelevant for vkQuake specifically, and maybe not worth concerning the user with.

The "on" value does exactly what "quakespasm" did before. Anyone who previously had "Always run" set to "vanilla" will now see "Always run" set to "on", though if they toggle it back and forth, they'll be exactly on the "quakespasm" behavior.

The old "vanilla" option didn't work significantly differently (holding shift while moving with keyboard does nothing instead of slowing you down; arguably a bug). I think the option existed mostly so a player could use that option to make a config that was compatible with other source ports not supporting the cl_alwaysrun CVAR, but since vkQuake writes to its own config that's not read by other source ports it doesn't seem there's any value to keeping this.

### DIFF
--- a/Quake/menu.c
+++ b/Quake/menu.c
@@ -1350,14 +1350,6 @@ enum
 	GAME_OPTIONS_ITEMS
 };
 
-enum
-{
-	ALWAYSRUN_OFF = 0,
-	ALWAYSRUN_VANILLA,
-	ALWAYSRUN_QUAKESPASM,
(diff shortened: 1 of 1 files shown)


Novum:
Agreed

PR #649, opened 2023-03-12
Fix menu cursor movement

Macil:
This fixes two issues:
1. The cursor in the sound menu can be moved too far with the arrow keys.
2. The cursor could not be moved with the down key past the padding gap in the options menu or the video menu.

(not related to the issue I opened previously about mouse cursor stuff)

### DIFF
--- a/Quake/gl_vidsdl.c
+++ b/Quake/gl_vidsdl.c
@@ -3705,6 +3705,8 @@ void M_Video_Key (int key)
 	case K_UPARROW:
 		S_LocalSound ("misc/menu1.wav");
 		video_options_cursor--;
+		if (video_options_cursor == VID_OPT_PADDING)
+			video_options_cursor--;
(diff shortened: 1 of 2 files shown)


Quake/gl_vidsdl.c
```diff
@@ -3844,7 +3848,7 @@ void M_Video_Draw (cb_context_t *cbx)
 
 		M_Mouse_UpdateCursor (&video_options_cursor, 12, 400, y, 8, i);
 		if (video_options_cursor == VID_OPT_PADDING)
-			video_options_cursor = VID_OPT_VSYNC;
+			video_options_cursor--;
```
Macil, line 3851: (just tweaking this to be more consistent with the adjustment elsewhere and to be less dependent on the specific menu items)

Novum:
Sorry I didn't see your PR. I did basically the same fixes.

PR #650, opened 2023-03-12
Replace "Screen Size" slider with "Interface Detail" no-gun/none/standard/full option

Macil:
This PR replaces the game options menu "Screen Size" slider with an "Interface Detail" setting that has "No gun"/"None"/"Standard"/"Full" options. This is done to make these HUD settings more discoverable to users who might not expect that a specific section of a slider toggles HUD elements, and to be more familiar with people used to the Kex port. (I honestly had no idea that the "No gun" and "None" behaviors were supported in the Screen Size slider! I thought I had tried the slider fully in the past and never explored its whole range for surprises again.) This is inspired by the Kex port which has a "HUD Style" setting with "Off"/"Minimal"/"Standard"/"Full" options. (This PR doesn't exactly match that setup because "HUD Style" is too close to the naming of the existing "Interface Style" option here, and also "Interface Style" already has a "simple" option that accomplishes what Kex's HUD Style "minimal" value does.)

This PR does sacrifice the ability to shrink the viewport through the options menu, though the whole range of viewsize settings is still supported through the console. My assumption is that this feature isn't commonly purposefully used, and that it was more useful in the distant past before windowed mode was supported and you wanted to save performance by shrinking the resolution to a very small value without affecting the HUD. (The Kex port lacks this feature.)

Additionally, the "-" sizedown key is changed so that it only shrinks the HUD but can't go as far as shrinking the viewport. It seemed wrong to me to let a default keybind set setting values that aren't available in the options menu. This behavior can be overridden by setting "viewsize_allow_shrinking 1".

This PR is a bit of an opinionated change. I would understand if you weren't interested in it. Let me know if you might instead accept a PR that split the "Interface Detail" setting to its own CVAR so that it could exist simultaneously in the options menu alongside the existing "Screen Size" slider, which would be stripped of its statusbar-affecting behavior.

### DIFF
--- a/Quake/gl_screen.c
+++ b/Quake/gl_screen.c
@@ -94,6 +94,7 @@ cvar_t scr_usekfont = {"scr_usekfont", "0", CVAR_NONE}; // 2021 re-release
 cvar_t scr_style = {"scr_style", "0", CVAR_ARCHIVE};
 
 cvar_t scr_viewsize = {"viewsize", "100", CVAR_ARCHIVE};
+cvar_t scr_viewsize_allow_shrinking = {"viewsize_allow_shrinking", "0", CVAR_ARCHIVE};
 cvar_t scr_fov = {"fov", "90", CVAR_ARCHIVE}; // 10 - 170
 cvar_t scr_fov_adapt = {"fov_adapt", "1", CVAR_ARCHIVE};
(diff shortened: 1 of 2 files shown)


Novum:
The original intent of the view size was very likely for performance in the stone ages.

Novum:
I'll merge this but tune it some more

PR #664, opened 2023-03-14
Restore keyboard controls

temx:
The options stay removed from the menu and are not written to the config file (except mouse look, which is always written as if it was enabled, for better compatibility).

### DIFF
--- a/Quake/cl_input.c
+++ b/Quake/cl_input.c
@@ -50,9 +50,10 @@ state bit 2 is edge triggered on the down to up transition
 ===============================================================================
 */
 
+kbutton_t in_mlook = {.state = 1}, in_klook;
 kbutton_t in_left, in_right, in_forward, in_back;
 kbutton_t in_lookup, in_lookdown, in_moveleft, in_moveright;
-kbutton_t in_speed, in_use, in_jump, in_attack;
+kbutton_t in_strafe, in_speed, in_use, in_jump, in_attack;
 kbutton_t in_up, in_down;
 
 int in_impulse;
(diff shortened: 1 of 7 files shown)


Novum:
So I thought about this. What about if people have weird values in their config already? At least the GOG release came with some messed up default cfgs.

Also the driftpitch stuff does weird stuff even without having any of the cvars on. If I just moved forward in the start level it pointed the screen down when walking down the stairs etc.

Novum:
One more question, do you _need_ the mouse lock toggle for your use case? Could that just be removed?

temx:
I do not, but I think being able to use the original keyboard controls is a desirable feature.

Now it's initialized to 1 so it'll always start on even the first time, no matter what you had in your config.

I've tested the original GOG config. That one actually had mouse look, the problems are the bindings - A and D aren't bound to strafe, and removing these doesn't fix anything, they must be rebound manually.

Novum:
I just really don't think any player actually uses this. Kex Quake also removed all of it. It's just cruft.

sezero:
> I've tested the original GOG config. That one actually had mouse look, the problems are the bindings - A and D aren't bound to strafe, and removing these doesn't fix anything, they must be rebound manually.

Do we not have a default.cfg in vkquake.pak overriding the one in pak0.pak?

temx:
The rerelease has (in the menu) all of the following:
- turn left / right (commands are left in vkQuake, but they do nothing now)
- look up / down (commands are in vkQuake, and they work)
- center view (command removed from vkQuake)
- strafe (command removed from vkQuake but left in configs, so ALT+TAB/ALT+ENTER print `unknown command "+strafe"` now)

The problem with GOG is that it has a config (outside `pak0.pak`), it's just not a WASD one. The only way to fix that is to detect that specific config and ignore it. Either case the removals do nothing to fix that.

Novum:
Fine :)

PR #707, opened 2024-07-12
build x64 binaries first, same order as the win installers

j4reporting:
people do not read,  see 
https://www.slipseer.com/index.php?threads/the-immortal-lock.526/post-3595


Novum:
Is this supposed to fix the order of the download links in "Releases"? It won't do that, that is strictly alphabetic. I looked into this before.

Issue #721, opened 2024-08-15
Mouse cursor remains visible when using AppImage version.


robojerk:
**LOW EFFORT BUG REPORTS WILL BE CLOSED**

**Describe the bug**
Using wayland version of kwin, mouse cursor remains visible when using AppImage. I didn't try a non AppImage binary.

**To Reproduce**
Launch vkquake appimage using kwin or gamescope. I didn't try Mutter or other compositors.  I didnt try in X.

**Expected behavior**
Mouse cursor should be visible ONLY when in the menu, once a game starts it hides.

**Screenshots**
I tried but spectacle and other screenshot apps wouldn't capture the cursor.  It's there though.

**Desktop (please complete the following information):**
 - OS: Linux (Bazzite 40) 6.9.12-203.fsync.fc40.x86_64
 - GPU vendor and model
  Optimus Laptop
      AMD Ryzen 7 4800H (Renoir series AMD Radeon RX Vega 7)
      Nvidia RTX 2060 Mobile
 - Driver version
      Nvidia driver  560.31.02
      AMD Driver 24.1.5

**Mod**
In vanilla version of games and all mods

**Additional context**
When running under kwin the Plasma feature to grow mouse on mouse shake is entertaining.

sezero:
> introduced in this commit? [libsdl-org/SDL@c39f86c](https://github.com/libsdl-org/SDL/commit/c39f86c28a54c3f108c31a58ff610fe9508d03c4)
> 
> Ubuntu 20.04 doesn't support wayland. Maybe compiling on 22.04 could resolve this? Otherwise maybe downgrade SDL2 to 2.28.5 for linux binaries for now.

Report this to SDL please

Kontrabant:
Should be fixed upstream now.

robojerk:
> @robojerk could you pls try if this artifact fixes the mouse issue for you? https://github.com/j4reporting/vkQuake/actions/runs/10431663762
> 
> This appimage was created from the current vkQuake git master with latest SDL2 from the release-2.30.x branch. So it's not exactly 1.31.1.

No more mouse cursor :)

robojerk:
I quickly ran the new build as normal with and setting the SDL_VIDEODRIVER to wayland.  Mouse cursor is gone in both. :)

Thanks...

robojerk:
Mine seemed locked into 60

Also the cursor reappeared when setting SDL_VIDEODRIVER=wayland

robojerk:
[...]

PR #727, opened 2024-08-23
use  commit 7ca3d26  from SDL2-2.30.x  branch

j4reporting:
fix for #721

this should get rid of the visible mouse pointer in last AppImage release


vsonnier:
Thank you for your support on #721, btw.

PR #728, opened 2024-08-23
Switch building AppImages to Ubuntu 22.04

j4reporting:
In case  you want to switch the docker image from 20.04 to 22.04  at some point.
This is not needed right now.


vsonnier:
Why not ? thanks :)

PR #745, opened 2024-11-06
Fix the README sections

Perlence:
I noticed that the MinGW section was in the MacOS section, which seemed odd. It was moved from the Windows section in f9df8516d3027da854186c74d36243f2d11cb8ad.

I see two options: move MinGW back to the Windows section, or raise its heading level so that MinGW is on the same level as the Linux and MacOS sections. I've chosen the first one.

Regarding the heading levels, see the commit message.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -69,6 +69,26 @@ Prerequisites:
 Open the Visual Studio solution, `Windows\VisualStudio\vkquake.sln`, select the desired configuration and platform, then
 build the solution.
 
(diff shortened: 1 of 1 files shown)


vsonnier:
> I see two options: move MinGW back to the Windows section, or raise its heading level so that MinGW is on the same level as the Linux and MacOS sections. I've chosen the first one.

Your choice was fine, thanks.

Issue #756, opened 2025-01-01
Sound not working on Linux.


dipled:
I'm using Linux Mint 22 x86_64 and the Steam version of Quake.

I downloaded the vkquake-1.31.3_linux64.tar.gz file from the releases, extracted it inside my Quake folder under <path_to_steam>/steamapps/common/Quake, renamed the .pak files to lowercase and ran the AppImage.

Game runs well, but all sounds appear to be missing. I know the Steam version doesn't have the music files, but every single sound is missing.

Novum:
Sound is working perfectly fine, this is a system issue.

dipled:
I don't have SDL_AUDIODRIVER configured to anything, in fact there are no SDL envvars configured. Is this supposed to happen? lol

dipled:
I compiled vkQuake locally and it worked, though, somehow ninja kept yielding  `fatal error: SDL.h: No such file or directory` on several files. I changed all `SDL.h` to `SDL2/SDL.h` and it solved the compilation issue.

Now the binary runs with sound.

I don't know if my SDL2 installation is messed up, because the files I had to change all had 
`
#if defined(SDL_FRAMEWORK) || defined(NO_SDL_CONFIG)
 #include <SDL2/SDL.h>
 #else
 #include "SDL.h"
 #endif
`

and I changed it to 

`
#if defined(SDL_FRAMEWORK) || defined(NO_SDL_CONFIG)
 #include <SDL2/SDL.h>
 #else
 #include "SDL2/SDL.h"
 #endif
`

For context, I didn't have SDL installed before and I installed it using the `libsld2-dev` package from apt.

Anyways, now I'm wondering why these variables aren't defined when they should be -_-

dipled:
Yes, I had ran the reconfigure and even the wipe commands, but ninja still would yield the same error.

Game is running with pulse
```
SDL audio spec  : 44100 Hz, 1024 samples, 2 channels
SDL audio driver: pulseaudio - Built-in Audio Analog Stereo, 65536 bytes buffer
Audio: 16 bit, stereo, 44100 Hz
```

dipled:
Yes, the output of sdl2-config matches the pkgconfig file content.

By the way, running the AppImage from terminal yields this

```
Couldn't init SDL audio: Failed loading libpulse.so.0: /lib/x86_64-linux-gnu/libsndfile.so.1: undefined symbol: mpg123_info2
Failed initializing sound
```

dipled:
[...]

PR #757, opened 2025-01-04
Keep building linux AppImage with libmad, resolves issue #756

j4reporting:
introduced  with commit 8b42de7  and  commit 7068875 
this breaks audio for newer linux systems If pulseaudio ( libpulse0 ) is used by the system. 

This can probably be reverted when we update the docker image to Ubuntu 22.04 at some point.

### DIFF
--- a/Packaging/AppImage/docker/Dockerfile
+++ b/Packaging/AppImage/docker/Dockerfile
@@ -16,7 +16,7 @@ RUN apt-get update \
 	libaudiofile-dev \
 	libdbus-1-dev \
 	libibus-1.0-dev \
-	libmpg123-dev \
+	libmad0-dev \
 	libopenal-dev \
 	libpulse-dev \
 	libsndio-dev \
--- a/Packaging/AppImage/run-in-docker.sh
+++ b/Packaging/AppImage/run-in-docker.sh
@@ -8,7 +8,7 @@ cd /usr/src/vkQuake
 
 rm -rf build/appimage
 
-python3 /opt/meson/meson.py build/appimage -Dbuildtype=release -Db_lto=true
+python3 /opt/meson/meson.py build/appimage -Dbuildtype=release -Db_lto=true -Dmp3_lib=mad
 ninja -C build/appimage
 
 cd Packaging/AppImage


Novum:
@vsonnier thoughts?

Novum:
If we want to keep 123 I think the correct fix is to include the old `.so` files from Ubuntu 20.04 that are incompatible in newer distros.

Axel-Reactor:
@j4reporting could you make a new PR which includes libpulse in the image? I don't really care about 2MB extra.

Novum:
@j4reporting Could you put up an alternative PR with libpulse and dependencies in the AppImage? I don't really think we care about 2MB extra if it solves the issue and we don't have divergence in what we ship.

vsonnier:
@Novum
> @vsonnier thoughts?
It really depends why we want 123 instead of mad? It's slightly faster and reads corrupt MP3s? I think we can live with mad for now.

Taking into account the .so Hell seen in #756 and this present PR I would simply put back `libmad` for Linux targets entirely, we have lived without `libmp123` for a long time after all. 
As @sezero said I followed @andrei-drexler on the change assuming it was because `libmp123` was maintained contrary to `libmad` ?

Now If we put back `libmad` for Linux targets only, we have 2 implementations to maintain although I do not expect to see MP3 bugs very often... 
On the other hand, I would like to keep the same code used as in QuakeSpasm and Ironwail for those major targets Windows and MacOS, so I would rather not revert `libmp123` for those.

So there is no ideal solution, so I'll let you choose, either this PR or reverting `libmp123` (as default) for Linux targets ?

vsonnier:
> for what's it worth, quakespasm still has mad as default option https://github.com/sezero/quakespasm/blob/271d588ef2d7d51328e0f349d1111f664c63cf6a/Quake/Makefile#L25-L26

I'm not attached to `libmp123`, so in that case we could also pretend the `libmp123` changes never occured and continue to use `libmad` for all, including Windows and MacOS. 

I don't see the supposed faster decoding for `libmp123` matters in any way, except on Rasberry PI-like targets ?

sezero:
>  it was because `libmp123` was maintained contrary to `libmad` ?

libmad is 'maintained' (ahem) at a different place for
some time: https://codeberg.org/tenacityteam/libmad
Fedora has switched to using it from that site for some
time, too.

However, libmad has been rock solid and doesn't need much
maintenance than what distros has been patching, and the
version I've been including in qs and other projects has
always been up to date.

>  so I'll let you choose, either this PR or reverting to `libmad` (as default) for Linux targets ?

Yup

> 

> > @j4reporting Could you put up an alternative PR with libpulse and dependencies in the AppImage? I don't really think we care about 2MB extra if it solves the issue and we don't have divergence in what we ship.
> 
> #758

I'm _really not_ fond of providing the appimage, but this? ughh. why not package the kernel, glibc and x11 along with others! ugh...

Novum:
> > it was because `libmp123` was maintained contrary to `libmad` ?
> 
> libmad is 'maintained' (ahem) at a different place for some time: https://codeberg.org/tenacityteam/libmad Fedora has switched to using it from that site for some time, too.
> 
> However, libmad has been rock solid and doesn't need much maintenance than what distros has been patching, and the version I've been including in qs and other projects has always been up to date.
> 
> > so I'll let you choose, either this PR or reverting to `libmad` (as default) for Linux targets ?
> 
> Yup
> 
> > 
> 
> > > @j4reporting Could you put up an alternative PR with libpulse and dependencies in the AppImage? I don't really think we care about 2MB extra if it solves the issue and we don't have divergence in what we ship.
> > 
> > 
> > #758
> 
> I'm _really not_ fond of providing the appimage, but this? ughh. why not package the kernel, glibc and x11 along with others! ugh...
Only stable interface on Linux is the kernel. No one else gives a f***.

Axel-Reactor:
I still don't understand why libmpg123 loads pulse?

sezero:
> I still don't understand why libmpg123 loads pulse?

No: libpulse loads libsndfile, libsndfile loads libmpg123

Axel-Reactor:
I see. Okay, let's use mad for AppImages for now until we can get to a newer Ubuntu. You're right.

Issue #698, opened 2024-06-22
Modern HUD all ammo counter


Zaratul555:
Like in the title. Its good to have modern hud but please implement all ammo counter from Iron Wail too.

vsonnier:
Not interested myself, but could welcome a contribution provided it doesn't break any of the existing CSQC Hud code.

Zaratul555:
Iron Wail had this and its not breaking anything. Also we got mod option for HUD now. Its not just copy paste code from IW?

![start_2024-06-28_23-35-42](https://github.com/Novum/vkQuake/assets/70495927/e9c3837a-c077-42f5-89ac-480026ff52ce)

Zaratul555:
@andrei-drexler
I see that You are adding commits to VKQuake. Its possible to add this feature by You?

andrei-drexler:
I have very little time/energy lately, and 
I think my time/energy would be better spent improving IW rather than reimplementing features that already exist in a different project.

andrei-drexler:
I have very little time/energy lately, and I think it would be better spent improving IW rather than reimplementing its features. That's where the commits you are seeing are coming from (the code being open-source and all), but it isn't me that's adding them here.

Zaratul555:
Obviously You are busy man with all of the coding and i respect Your decision. Its just wierd that Iron Wail HUD features has been imported only partially.

vsonnier:
Completed by #842.

PR #764, opened 2025-01-14
align modern hud closer to the remaster and render  keys for hipnotic as well

j4reporting:
The minimal modern hud  does always render armor and keys  on kex engine.
There is no difference between viewsize 100  and 110, unless we  port the inventory from IW.


vsonnier:
I'll try to look at this Soon (TM)

Novum:
I didn't really spend a lot of time on the HUD stuff back then, there are many improvements that could be made.

vsonnier:
Also related to https://github.com/Novum/vkQuake/issues/698. At one point I would like to immplement the set of HUDs, from Ironwail and get on with it, but I'm not super-interested in the matter and lacks time.

Issue #761, opened 2025-01-07
Glow maps don't appear to work correctly on md5 models


NurglesHogweed:
When I try to use a glow map, the whole model gets fullbright instead of the opaque pixels only. To reproduce, one can use [q2m](https://www.moddb.com/mods/q2m-quake-2-weapons-for-quake) 1.12 and files from the attached archive. I have updated VkQuake to the latest master revision.
![glow](https://github.com/user-attachments/assets/b07f9511-6d8f-4ca4-9b6d-0d14c898ed0f)
[shotg.zip](https://github.com/user-attachments/files/18338364/shotg.zip)

vsonnier:
Hi @NurglesHogweed thanks for your report. I think I've fixed it, my emulating for RGBA textures what is done with INDEXED (palette) using an external palette.

The glow texture pixels  are pre-processed as this:
- if alpha > 0 , this is a "fullbright" pixel => forced alpha = 255
- if alpha = 0 (like in this glow texture), this is the 'mask' => color forced to black (0,0,0) and alpha = 255 as well.

PR #765, opened 2025-01-18
Standalone md5 model support

NHogweed:
Now, if there is a md5 model, but no mdl with the same name, the engine will use the former instead of returning an error. This ignores r_loadmd5models and r_md5models.

### DIFF
--- a/Quake/gl_model.c
+++ b/Quake/gl_model.c
@@ -141,6 +141,8 @@ void *Mod_Extradata_CheckSkin (qmodel_t *mod, int skinnum)
 			return mod->extradata[1];
 		if (r_md5models.value && mod->md5_prio && skinnum < ((aliashdr_t *)mod->extradata[1])->numskins)
 			return mod->extradata[1];
+		if (!mod->extradata[0])
+			return mod->extradata[1];
 	}
 	return mod->extradata[0];
 }
@@ -492,9 +494,10 @@ static qmodel_t *Mod_LoadModel (qmodel_t *mod, qboolean crash)
 	//
 	// load the file
 	//
+	mod_type = COM_FileExists (mod->name, &mod->path_id);
 	qboolean	 md5_loaded = false;
 	unsigned int md5_path_id = 0;
-	if (r_loadmd5models.value)
+	if (r_loadmd5models.value || !mod_type)
 	{
 		char newname[MAX_QPATH];
 		q_strlcpy (newname, mod->name, sizeof (newname));
@@ -506,6 +509,8 @@ static qmodel_t *Mod_LoadModel (qmodel_t *mod, qboolean crash)
 			if (buf)
 			{
 				Mod_LoadMD5MeshModel (mod, buf);
+				if (!mod_type)
+					goto md5_forced;
 				md5_loaded = true;
 				md5_path_id = mod->path_id;
 			}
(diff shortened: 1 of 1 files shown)


vsonnier:
I @NHogweed, thanks for this. Now, the code around the selection of which model to load it getting too messy for my taste, especially with the `goto` addition :) I'm going to move the model selection to a separate function to be more explicit and readable.

No need to touble yourself further though, I'll take care of the rest from this.

NHogweed:
OK. I'm keeping my code until that though, just for any case.

vsonnier:
Hi @NHogweed finally I merged your modification as-is. I wanted to be pedantic at first and redo this, but hey, I'm lacking time and as we say home : "le mieux est l'ennemi du bien" or if you prefer here Practicality beats purity" :)

Seriously though, your change was not that big, and is still easy enough to follow.

Thanks for your contribution !

NHogweed:
I think it would be better for compatibility (at least with FTEQW and QSS) purposes to revert my changes to `Mod_LoadModel` (addition to `Mod_Extradata_CheckSkin` is still needed) and add lines 452 - 454 from [QSS](https://github.com/Shpoike/Quakespasm/blob/qsrebase/Quake/gl_model.c#L451) instead. The normal way to use standalone md5meshes in other engines is to refer to them as themselves in QuakeC. They can be referred to as mdls, but this is cvar-dependent.

vsonnier:
Please provide another PR, because I don't see how the changes you mention could work and still be compatible with the 
```c
if (r_loadmd5models.value) {
... 
}
```
branch ?

NHogweed:
[Let's compare gl_model.c right before the latest change and what I propose now.](https://github.com/Novum/vkQuake/compare/1.32.0...NHogweed:vkQuake:master#diff-dd00f9a260a231c561c39df018db602d64a2ee4a68f4220ffc4687595720681e)

NHogweed:
> this last change do not work for MD5 standalone anymore

This is not what I saw after its compilation.
> You need to protect against loading MD5 twice

Really? As far as I understand,
`char newname[MAX_QPATH];`
`q_strlcpy (newname, mod->name, sizeof (newname));`
`char *extension = (char *)COM_FileGetExtension (newname);`
`if (strcmp (extension, "mdl") == 0)`
checks if the loaded file has .mdl extension. And
`case (('M'<<0)+('D'<<8)+('5'<<16)+('V'<<24)):`
`		Mod_LoadMD5MeshModel(mod, buf);`
is a check for the loaded file's content. So md5 replacement won't be loaded for anything but a .mdl file, and if it is indeed an alias model, the corresponding option in the `switch (mod_type)` section will be used. I can imagine double md5 loading if md5's extension was changed to .mdl, but I doubt this is a reasonable thing to do.

vsonnier:
> Just tried [RME](https://www.moddb.com/mods/quake-remastered-models-extended-quake-rme/addons/quake-rme) for hipnotic, and it worked as usual.

Can't be, the MD5 of this package are invalid, ( https://github.com/andrei-drexler/ironwail/issues/356#issuecomment-2469667005 ) unless you have a particular version ? 
They still crash on me, copying contents into QuakeEnhanced directory (a.k.a Remaster from GOG)

Now for the code, I think you are right:

The switch(mod_type) is used to dispatch to the appropriate loader but indeed `md5_loaded` can only be true if `mod->name` was a .mdl, so indeed we call Mod_LoadMD5MeshModel() at that place only for a true MD5, that was never called in the .mdl replacement option.
To be clearer, I'll rename   `md5_loaded` into  `md5_replacement_loaded`.

Following that logic, you proposed change should bring MD5 standalone support. 

The previous code `md5_forced` allowed to load a standalone MD5 (without .mdl companion) while in QuakeC the specified model was .mdl. 

That's how I tested unsing Q2M (https://github.com/Novum/vkQuake/issues/761) but this is not a valid situation, the .mdl is supposed to be provided if specified in QuakeC, and for MD5 it is the same.

vsonnier:
Thanks, it works, added in commit https://github.com/Novum/vkQuake/commit/8dfa015a9d188629e1bf26dd46d10912c7dcc7d5

> Maybe it's a good idea to rename md5-related cvars too?

Yes `r_loadmd5models` was replaced in `r_allow_replacement_md5models`, default to 1 but NOT saved, because by default we want replacement possible, AND in fact this is only taken into account at each new game / level, so changing it dynamically do not have the expected effect. So let's consider this as a "debug" feature.

`r_md5models` stays the same and is controlled by Models > Remastered / Classic.

PR #772, opened 2025-02-03
activate  renderer/network isolation with host_maxfps >= 73

j4reporting:


vsonnier:
Hi @j4reporting thanks for the change. On my machine I don't see much of a change compared to `master` in the Immortal Lock, but I don't see a loss of performance either, so we not taking it ? For the Immortal Lock though, `host_maxfps 0` stay unplayable no matter the version.

vsonnier:
@j4reporting Indeed I see a big difference using `host_frametime = q_max(accumtime, host_netinterval)` : with this `hasit_maxfps 0` is playable with the Immortal Lock save, and the `fps` seems now more or less constant whatever `host_maxfps` is 0, 58 , 72, 73 or whatever. 

So, I've commited the change in https://github.com/Novum/vkQuake/commit/60957d171bfe546cbe0026c2097a251cc2ec092a

Issue #590, opened 2022-12-10
macOS X packaging


Novum:
Would be nice to have an actual macOS release, but it's quite involved. Maybe we should port the QuakeSpasm XCode project.

Also don't really want to pay for a dev account to sign it. Without that M1 macs refuse to run the exe without the right click->open workaround that not everyone knows about.

sezero:
Maybe @ericwa can help

Novum:
Looking at TrenchBroom downloads from github the distribution is about 80% Windows 15% Linux and 5% macOS.

Not sure how that translates to actual players though.

johnalanwoods:
I'm happy to help with this, and have a macOS developer account, it runs terrifically on my M1 Pro, 120FPS everything maxed out.

> Apple probably will remove OpenGL at some point.

This wouldn't affect this project, right?

Novum:
> This wouldn't affect this project, right?
That's the point. QuakeSpasm and other ports would stop working.

Novum:
I did some amount of work here:
https://github.com/Novum/vkQuake/commit/1c2ddd90041fd277a5f2a37452e5d95858309565
This will produce a universal binary.

But I'm not sure if that's actually the right way to approach this, probably want an XCode project to get all the app packaging etc. Doing that manually is a pain.

karstenBriksoft:
I don’t know the details of how they build, but MacSourcePorts has a fork (https://github.com/MacSourcePorts/vkQuake) and also the latest build that is ready for download at: https://macsourceports.com/game/quake

Novum:
Their scripts might be useful. Don't understand why they don't want to upstream changes.

temx:
It looks like the script they actually use is at https://github.com/MacSourcePorts/MSPBuildSystem/tree/main/vkquake

Novum:
Yeah, the interesting stuff is in common: https://github.com/MacSourcePorts/MSPBuildSystem/tree/main/common

tomkidd:
Howdy, I'm the one who runs Mac Source Ports. 

Back when I started out (around Sept. 2021) I did a fork of vkQuake because I needed to tweak the then-Makefile to do cross compilation but at some point after the move to Meson, I was able to just use the repo's contents directly, no more need to modify anything. So that's why that fork I did is so behind (I should probably hide or delete it).

[...]

PR #776, opened 2025-02-12
README note on releases for Windows, Linux and macOS

stared:
I added a note where releases can be downloaded from.

Previously, README had no notes on Windows and Linux binaries.
Also, it was a bit confusing that while the MacOS tests are passing, there are no MacOS binaries, vide #590.


vsonnier:
Thanks !

PR #784, opened 2025-03-19
Import DP_SV_NODRAWTOCLIENT and DP_SV_DRAWONLYTOCLIENT from QSS

NHogweed:
The [attached archive](https://github.com/user-attachments/files/19351367/test_qcext.zip) contains a compiled .exe and a QuakeC build using these extensions. To test them, one can start a coop game and then join it from another instance of vkQuake which can be launched even on the same computer.

How to trigger:
[DP_SV_NODRAWTOCLIENT](https://quakewiki.org/wiki/DP_SV_NODRAWTOCLIENT) - when a monster is hit by an axe, it will become invisible to the attacker.
[DP_SV_DRAWONLYTOCLIENT](https://quakewiki.org/wiki/DP_SV_DRAWONLYTOCLIENT) - when a player's armor is > 0, his/her legs (from [Quake 1.5](https://www.moddb.com/mods/quake-15)) become invisible to other clients.

### DIFF
--- a/Quake/progs.h
+++ b/Quake/progs.h
@@ -279,8 +279,8 @@ struct pr_extfields_s
 	/*ssqc-only*/                                                                   \
 	QCEXTFIELD (items2, "//.float")									 /*float*/      \
 	QCEXTFIELD (movement, ".vector")								 /*vector*/     \
-	QCEXTFIELD (viewmodelforclient, ".entity")						 /*entity*/     \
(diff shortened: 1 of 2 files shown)


vsonnier:
Thanks ! I have no way to test your changes though, so I'm trusting you on this. I've commit your change manually addinf Clang format fixes and some more comments.

PR #786, opened 2025-03-21
Add nix-shell with shell.nix

kugodd:
Hello,
I don't contributes a lot, but here a small for devs that use [Nix](https://nixos.org/).
Have a nice day,
KuGodd


kugodd:
Sorry, yes this is to be able easily build with nix, which helps to get consistent package during development.

PR #792, opened 2025-05-01
Detach worker threads

bubbleguuum:
Worker threads are created and never referenced.
They can be safely detached with SDL_DetachThread.

This fixes these warnings displayed upon exiting vkquake when run using sdl2-compat (SDL2 compatibility layer that uses SDL3 behind the scenes) used by some Linux distros:

INFO: Leaked thread (0x55ad3e03cc20)
(repeated for each thread)

### DIFF
--- a/Quake/tasks.c
+++ b/Quake/tasks.c
@@ -92,7 +92,6 @@ typedef struct
 } task_counter_t;
 
 static int					 num_workers = 0;
-static SDL_Thread		   **worker_threads;
 static task_t				 tasks[MAX_PENDING_TASKS];
 static task_queue_t			*free_task_queue;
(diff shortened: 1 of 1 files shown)


vsonnier:
@bubbleguuum Thanks, looks like it works on my side (Win10 x64) as for Linux I'll have to trust you.

bubbleguuum:
Confirming I tested on Linux, on openSUSE Tumbleweed using [sdl2-compat](https://github.com/libsdl-org/sdl2-compat) meaning SDL3. SDL3 now displays "Leaked thread" warnings on SDL_Quit() for any non-detached thread that has not terminated.

vsonnier:
This is good to go then

PR #794, opened 2025-05-08
Split update_lightmap shaders in two groups

### LINKED ISSUE #793: Miss-rendering when running from raspberry pi 5
When launching the game in raspberry pi 5, there are missrenderings in the game.

Image

When running the game through Vulkan ValidationLayers, a validation warning appears:

```
Validation Warning: [ Undefined-Value-StorageImage-FormatMismatch-ImageView ] | MessageID = 0x13365b2
vkCmdDispatch(): the storage image descriptor [VkDescriptorSet 0x4620000000462, Set 0, Binding 0, Index 0] is accessed by a OpTypeImage that has a Format operand Rgba8 (equivalent to VK_FORMAT_R8G8B8A8_UNORM) which doesn't match the VkImageView 0x4580000000458 format (VK_FORMAT_A2B10G10R10_UNORM_PACK32). Any loads or stores with the variable will produce undefined values to the whole image (not just the texel being accessed). While the formats are compatible, Storage Images must exactly match. Few ways to resolve this are
1. Set your ImageView to VK_FORMAT_R8G8B8A8_UNORM and swizzle the values in the shader to match the desired results.
2. Use the Unknown format in your shader (will need the widely supported shaderStorageImageWriteWithoutFormat feature)
Spec information at https://docs.vulkan.org/spec/latest/chapters/textures.html#textures-format-validation
Objects: 4
    [0] VkCommandBuffer 0x432fb7a0
    [1] VkPipeline 0x1130000000113
    [2] VkDescriptorSet 0x4620000000462
    [3] VkImageView 0x4580000000458
```

The issue seems about a shader using a rgba8 format when the color _format  selected is rgb10_a2. This doesn't happens if we disable the latter in the driver, as the game defaults to use a rgba8 color_format.

The solution seems to do like in other cases: have a set of shaders to use with 10bit color formats, and other set to use with 8bit color_format.

jasuarez:
One of the group uses an rgb10_a2 output_image and the other uses a rgba8 output_image. The first group will be used when the driver supports VK_FORMAT_A2B10G10R10_UNORM_PACK32 color format, while the second will be used in other case.

This fixes #793.


vsonnier:
Thanks @jasuarez for your contribution ! LGTM I've only tested on Win10, I'll trust you for the rest.

PR #821, opened 2025-11-07
vkQuake embedded pack : removed dependency on compressor tooling

vsonnier:
- Removed zopfli external dependency, also because discontinued
- replaced by bintoc -c option, built around our existing miniz
- Upgraded to miniz 3.10 to bring back compression routines, was stripped previously
- Embed miniz.c stb-style (supported by miniz) dicrectly into common.c and bintoc.c compilations units
- Visual Studio projects, Makefiles and Meson now rebuild both vkquake.pak and embedded_pak.c from scratch.

That took a fuckton of time to tune, so this is as good it gets from me. Fell free to submit PR, but I won't touch the subject anytime soon myself.

Only 64bit targets has been changed, because 32bit will disappear very very soon, see #815.


vsonnier:
Hi @j4reporting On the contrary, I would like to have less choices for building options, not more. Did I ever say I  __hate__ build systems ? :) 

Actually the only reason `embedded_pak.c` is still in Git is because I didn't manage (or spent even more time looking for a solution) with either Make or Meson, to indicate it was a generated file so didn't need to be present at the begining of the build, while still listed for compilation.

I think there is no harm that binary resources are rebuild from scratch each time (just like code), especially when the whole vkQuake take 10 seconds to build fully on a very modest machine.

Issue #804, opened 2025-08-21
Feature request: Windows Arm64 build


teobugslayer:
As title says, it would be awesome if we can use a native Arm64 build of vkQuake.

Sadly, this probably will require at minimum moving from SDL2 to SDL3, because the latter has a supported build for WoA.

teobugslayer:
No, it could not start. The error message is:

---------------------------
vkQuake.exe - Application Error
---------------------------
The application was unable to start correctly (0xc000007b). Click OK to close the application. 
---------------------------
OK   
---------------------------

There's nothing more in the Event log, just the same message logged. 

I tested with the GOG distribution (Quake - the offering).

WinDbg !analyze says:

Event is not an exception - analysis may be incorrect

PROCESS_NAME:  .exe

ERROR_CODE: (NTSTATUS) 0xc000007b - {Bad Image}  %hs is either not designed to run on Windows or it contains an error. Try installing the program again using the original installation media or contact your system administrator or the software vendor for support. Error status 0x

SYMBOL_NAME:  ntdll!NtWaitForWorkViaWorkerFactory+4

MODULE_NAME: ntdll

IMAGE_NAME:  ntdll.dll

FAILURE_BUCKET_ID:  INVALID_IMAGE_FORMAT_c000007b_ntdll.dll!NtWaitForWorkViaWorkerFactory

teobugslayer:
The same scene renders at ~120 fps. I would say slightly lower but not noticeable (I am a very casual player).

vsonnier:
Hi @j4reporting since it seems to work, (thanks !) is https://github.com/Novum/vkQuake/compare/master...j4reporting:vkQuake:clangarm64 ready for a PR ?

teobugslayer:
I'd be happy to test it for you, just at-mention me, so that I receive a notification.

vsonnier:
the corresponding PR has been merged.

PR #826, opened 2025-11-10
add windows arm64 binaries 

j4reporting:
Uses  MSYS2 clangarm64 env   on Win11 arm64

### DIFF
--- a/.github/workflows/build-msys2-clangarm64.yml
+++ b/.github/workflows/build-msys2-clangarm64.yml
@@ -1,0 +1,52 @@
+name: Windows arm64 CI (mingw)
+
+on: [push, pull_request]
+
+jobs:
+  build-mingw:
+    name: ${{ matrix.platform.name }}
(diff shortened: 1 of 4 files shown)


sezero:
@j4reporting: Are the windows arm64 builds run-tested?

I plan on adding windows arm64 support to quakespasm, with prebuilt SDL and audio codec dlls. Would you be able to test when/if I do so?

j4reporting:
@sezero,   I don't have  any win/arm64 hw.  Only  the initial build was tested by @teobugslayer,  see  https://github.com/Novum/vkQuake/issues/804 

That is why the release notes include  the following line :)
>Please note: Windows arm64 binaries are not tested / not supported and are provided "as-is".

teobugslayer:
@sezero I test every vkQuake arm64 build on my x1 laptop and so far they worked flawlessly. But this is not official, nor exhaustive - I just drop the files from the latest build in my GoG quake distribution folder and move around a few AD levels.

PR #828, opened 2025-11-11
[Feature] MD3 Support

vsonnier:
The discussion here : https://github.com/andrei-drexler/ironwail/issues/444 convinced me to try it on vkQuake. 

The idea is to run the rendering for both MDL and MD3 on the same `alias.vert` pipeline because both are vertex-animated meshes and the rendering seems universal enough, without too much changes. (as my limited Vulkan understands it)

I used mostly QSS MD3 loading code as a starting point. 

Otherwise, I'll follow the texture conventions and file extensions set by the matching Ironwail PR: https://github.com/andrei-drexler/ironwail/pull/451 , so not necessarily copying what QSS did.

Feature-wise, this PR aims to support the same as https://github.com/andrei-drexler/ironwail/pull/451 except for the `.skin` feature which is to be done later.  

For now, Surface-based and Model-based namings are supported, where S is surface number, X is skin number, Y is skingroup animation frame (all starting at 0), ext is the image format file extension (png, tga, jpg, pcx, lmp) :

- Surface-based naming : ( idem https://github.com/andrei-drexler/ironwail/pull/451)
  - surfacename_X_Y.ext 
  - surfacename_X.ext ( 0..X-1 skin, 1 framegroup)
  - surfacename.ext (1 skin, 1 framegroup)
  
- Model-based naming : either for a replacement of `model_name.mdl` or a native `model_name.md3`, searched if the surface-based naming fails. The advantage is to not depend on the surface name hardcoded in the MD3:

  - model_name.md3_S_X_Y.ext
  
In addition I'm hardcoding the priority:  MD3 first (to be able to override Remastered models), then MD5, then MDL, instead of IW solution to parametrize that order by yet another CVAR and more supporting code. 

Loading priority is only usefull for designers anyway, and there are already `r_allow_replacement_md5models` and `r_allow_replacement_md3models` to selectively disable loading for debuging purposes. (changes applied at level loading only).

Finally the effective loaded models can be displayed by the `mcache` console command, as before.

### DIFF
--- a/Quake/common.c
+++ b/Quake/common.c
@@ -1599,13 +1599,12 @@ void COM_Init (void)
 ============
 va
 
-does a varargs printf into a temp buffer. cycles between
-4 different static buffers. the number of buffers cycled
-is defined in VA_NUM_BUFFS.
+does a varargs printf into a temp buffer. Cycles between
+VA_NUM_BUFFS different static buffers.
 FIXME: make this buffer size safe someday
 ============
 */
-#define VA_NUM_BUFFS 4
+#define VA_NUM_BUFFS 8
 #if (MAX_OSPATH >= 1024)
 #define VA_BUFFERLEN MAX_OSPATH
 #else
--- a/Quake/gl_model.c
+++ b/Quake/gl_model.c
@@ -48,7 +48,7 @@ cvar_t r_allow_replacement_md5models = {"r_allow_replacement_md5models", "1", CV
 // r_allow_replacement_md3models = 1 allow loading of MD3 replacement models if available, 0 to forbid it for debug purposes.
 cvar_t r_allow_replacement_md3models = {"r_allow_replacement_md3models", "1", CVAR_NONE};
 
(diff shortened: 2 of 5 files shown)


vsonnier:
@j4reporting look at the fullbrights textures for them 😁this is creative experiment of mine. Soldier texture has been made semi transparent for testing as well. Load the Necropolis save to see a regular Demon.

j4reporting:
There are no duplicated soldiers.  just   `skill 2    + notarget`  i guess :)
Sorry should have mention that I did mean the transparent soldiers but looking at the skin files makes sense now. 

 can't load the savegames though.  What mod or additional pakfiles do you have installed?
```
loading game  md3_TEST/s1.sav...
'axeonly' is not a global
'savedgamecfg' is not a global
'startingserverflags' is not a global
'reset_flag' is not a global
Can't find function monster_touch
Host_Error: ED_ParseEdict: parse error
```

PR #858, opened 2026-01-18
Better dynamic shadow quality & dynamic enemy shadows

Novum:


j4reporting:
@novum, `r_rtshadows > 0`  rockets  no longer illuminate their surroundings. The same applies to lava balls (i.e. start map).   Only when a lava ball falls back into the lava is the surrounding area illuminated for a split second.
Regression or deliberate change?
tested only on linux / nvidia   current  drivers 590.48.01

PR #874, opened 2026-02-26
Paranoid instrumentation, some fixes + built-in stack traces

vsonnier:
- Fixes for edicts, physics, overflows... + simplifications/clarifications about edicts management
  
- Tasks : moved platform-specific code to Sys
 
- Added Sys_StackTrace and use it in Host_Error/Sys_Error:

Windows MSVC: add DbgHelp lib dependency + package *.pdb in zips and Installers
Windows MSYS2 : .pdb support plainly broken or too complicated, generate DWARF debug info in the binary that can be used by addr2line post-mortem.
      - x86_64 : addr2line works, after much trial and error...
      - ARM64  : should work the same, Windows API calls are supposed to be universal => not tested.
     -  Linux/MacOS/FreeBSD...etc : always build with -g -fno-omit-frame-pointer, never strip binaries => not tested, Makefile _should_ work, __but__  CI Meson builds looks rally too small to really include degug info...

### DIFF
--- a/.github/workflows/build-mac.yml
+++ b/.github/workflows/build-mac.yml
@@ -15,7 +15,7 @@ jobs:
     - name: Install Dependencies
       run: brew install molten-vk vulkan-headers glslang spirv-tools sdl2 libvorbis flac opus opusfile flac mpg123 meson
     - name: Build vkQuake
-      run: meson build --buildtype=${{ matrix.configuration }} && ninja -C build
+      run: meson build --buildtype=${{ matrix.configuration }} -Ddebug=true -Dstrip=false && ninja -C build
     - name: Upload vkQuake
       uses: actions/upload-artifact@v4
       with:
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -37,3 +37,4 @@ jobs:
         path: |
             Windows\VisualStudio\Build-vkQuake\**\*.exe
             Windows\VisualStudio\Build-vkQuake\**\*.dll
+            Windows\VisualStudio\Build-vkQuake\**\vkQuake.pdb
--- a/Packaging/AppImage/run-in-docker.sh
+++ b/Packaging/AppImage/run-in-docker.sh
@@ -8,7 +8,7 @@ cd /usr/src/vkQuake
 
 rm -rf build/appimage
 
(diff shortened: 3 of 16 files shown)


j4reporting:
> but CI Meson builds looks rally too small to really include degug info... 

I think binary is  probably stripped by  linuxdeploy

Novum:
I do not like this change at all. I don't like getting a window pop up in the debugger and then having to hunt for the call stack because it's not the same thread. Not ok.

Also why is this enabled even in release builds apparently?

Novum:
If that's the case, we probably should just disable the pop up in debug builds honestly. Debug is only for development and you can attach a debugger if you need call stacks?

vsonnier:
I'm replacing the PARNOID builds (I didn't coin it, it was there before, I just expanded the coverage) with regular Debug builds so it can still be useful for users to try and report errors including call stacks

However there is indeed APIs to detect (both on Windows and Linux) it we are runing in a debugger so I can add it to `Sys_Error` and bypass pop-ups and stack traces while debugging.

j4reporting:
so for some reason  stack traces  provided by the ubuntu20lts/gcc9  toolchain ;ook different and you have to use the offsets/addresses in the parantheses!  The addresses in the squake brackets are  incorrect here.

latest git master:
```
STACK TRACE:
0x55f25888a73a 0x55f25888abf4 0x55f25888e7c6 0x55f2587b8fde 0x55f2587461c2 0x7f737a8105b5 0x7f737a810668 0x55f2587462ee 
0 : ./vkquake.AppImage(+0x16373a) [0x55f25888a73a]
1 : ./vkquake.AppImage(+0x163bf4) [0x55f25888abf4]
2 : ./vkquake.AppImage(+0x1677c6) [0x55f25888e7c6]
3 : ./vkquake.AppImage(+0x91fde) [0x55f2587b8fde]
4 : ./vkquake.AppImage(+0x1f1c2) [0x55f2587461c2]
```
extract the  binary from appimage archive  and use that for addr2line 
```
$ ./vkquake.AppImage --appimage-extract
[deleted]

$ addr2line -e squashfs-root/usr/bin/vkquake 0x1677c6
/usr/src/vkQuake/build/appimage/../../Quake/wad.c:82

$ addr2line -e squashfs-root/usr/bin/vkquake 0x55e69c8c57c6
??:0
```

PR #893, opened 2026-04-12
Merge workflows and update issue template

Darvisim:
* Merged all workflow files like build-linux.yml, build-macos.yml and others into a single workflow file main.yml.
* Updated issue templates by replacing the .md files with the .yml files.


vsonnier:
> Merged all workflow files like build-linux.yml, build-macos.yml and others into a single workflow file main.yml.

Please explain the rationale behind this, for me it does not make the maintenance any easier. Platforms are different enough to deserve a different config. @Novum thoughts ?

Darvisim:
> > Merged all workflow files like build-linux.yml, build-macos.yml and others into a single workflow file main.yml.
> 
> 
> 
> Please explain the rationale behind this, for me it does not make the maintenance any easier. Platforms are different enough to deserve a different config. @Novum thoughts ?

In my opinion, it is more convenient to configure the platforms via their dedicated jobs rather than separate workflows. Hence, I merged them into a single file. However, I am uncertain how this would affect maintenance. Wouldn't it be more efficient to maintain one workflow file compared to maintaining workflow files for each platform?

vsonnier:
Sorry the `.yaml` bug template format do not work, and I'm not intrested in the other `main.yml`. And frankly this subject do not interest me enough compared to the free time I have.  Thanks anyway.

PR #897, opened 2026-04-22
update linuxdeploy 

### LINKED ISSUE #896: AppImage no longer works on Fedora due to FUSE 2
**Describe the bug**
Trying to run the appimage on Fedora 44 produces this error:
```
dlopen(): error loading libfuse.so.2

AppImages require FUSE to run. 
You might still be able to extract the contents of this AppImage 
if you run it with the --appimage-extract option. 
See https://github.com/AppImage/AppImageKit/wiki/FUSE 
for more information
```
It needs to be updated to a newer appimage runtime that includes FUSE 3.

**To Reproduce**
Try and run the appimage on Fedora 44. Launching from a gui environment will seem to do nothing at all. Launching from the command line will produce the error above.

**Expected behavior**
Launch instead of erroring out.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Fedora 44
 - GPU vendor and model: does not matter for this issue
 - Driver version: does not matter for this issue

j4reporting:
to https://github.com/linuxdeploy/linuxdeploy/releases/tag/1-alpha-20251107-1
version includes support for  FUSE3

resolves  https://github.com/Novum/vkQuake/issues/896


j4reporting:
@vsonnier  here is an updated  1.34.1  linux appimage  
https://github.com/j4reporting/vkQuake/actions/runs/24766520288

This build  should probably replace  the current  fuse2 only  1.34.1 appimage.

PR #899, opened 2026-04-22
add missing check for ENTALPHA_DEFAULT in R_IsEntityTransparent

### LINKED ISSUE #767: lavaalpha issues in ne_ruins (invisible lava)
In the "Altar of Storms" mod:
https://www.quaddicted.com/reviews/ne_ruins.html

At some point during the second map the lava rises and floods the lower part of the level. It is invisible in vkQuake.

I noticed this bug just as vkQuake was archived nearly two years ago so I couldn't create an issue here but then I was reminded of this bug today.
The problem appeared around the time Temx reworked the engines' alphasorting, the mod was working fine in vkQuake in older versions.

I have tried disabling r_alphasort but it doesn't change the problem.
I noticed that r_wateralpha is at 0.7 and r_lavaalpha is at 0.
Changing either r_wateralpha or r_lavaalpha to 1 makes the lava appear, but any other value (between ==0 and <1) the lava is invisible.
I don't know if the issue is caused because the lava is invisible at the start of the map or what (as it rises during the level).

I use vkQuake in Win10 with a geforce980

In vkQuake:
Image
In Ironwail for comparison:
Image

A save to the spot with the problem:
[save.zip](https://github.com/user-attachments/files/18509826/save.zip)

j4reporting:
add missing check for ENTALPHA_DEFAULT
  otherwise,  water entities with a fixed alpha value of 1.0f might inherit the
  current value of r_wateralpha < 1.0  and be consodered transparent.

resolves  https://github.com/Novum/vkQuake/issues/767

### DIFF
--- a/Quake/gl_rmain.c
+++ b/Quake/gl_rmain.c
@@ -444,7 +444,7 @@ static qboolean R_IsEntityTransparent (entity_t *e, qboolean *opaque_with_transp
 {
 	qboolean transparent = ENTALPHA_DECODE (e->alpha) != 1;
 	*opaque_with_transparent_water =
-		(!transparent && e->model->type == mod_brush && e->model->used_specials & SURF_DRAWTURB &&
+		(!transparent && e->model->type == mod_brush && e->model->used_specials & SURF_DRAWTURB && e->alpha == ENTALPHA_DEFAULT &&
 		 ((e->model->used_specials & SURF_DRAWLAVA && (map_lavaalpha > 0 ? map_lavaalpha : map_fallbackalpha) != 1) ||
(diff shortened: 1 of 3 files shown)


vsonnier:
OMG it works ?  Thanks !

j4reporting:
You sound surprized.  :)    And all w/o the use of AI. 

The commit msg is somewhat  misleading, But `R_IsEntityTransparent()` would mark full opaque water  surfaces with fixed alpha 1  as transparent  when ` r_wateralpha < 1 ` and send it the wrong render path.  Ofc, the edict would  retain its alpha 1 value.
The "first bad commit"  exposed those surfaces.
We could also test for `e->alpha != ENTALPHA_ONE`  instead.

PR #901, opened 2026-05-02
WBOIT & classify brush textures by type on load

Novum:


Shaders/wboit.inc
```diff
@@ -0,0 +1,8 @@
+void WriteOITTransparency (vec4 frag_color, out vec4 oit_accum, out float oit_reveal)
+{
+	frag_color = clamp (frag_color, 0.0f, 1.0f);
+	float z = 1.0f / gl_FragCoord.w;
```
vsonnier, line 4: Probabably a supid question, but is this formula take into account we use https://github.com/Novum/vkQuake/pull/336 ?

PR #903, opened 2026-05-04
Use insert mode in console by default rather than overwrite mode

gera2342:


vsonnier:
Like Ironwail. Why not. Thanks @gera2342.

PR #908, opened 2026-05-14
Leaf and Backface culling batching

vsonnier:
In r_world.c `R_MarkSurfaces` the idea is to process leafs and surfaces not by batches of 32 but by splitting the whole bunch of them by the number of workers so that in case of a huge number of leafs/surfaces, we can gain some performance by making some bigger work items for tasks, assuming `R_MarkLeafsSIMD` `R_MarkLeafsParallel` `R_BackfaceCullSurfacesSIMD` `R_BackfaceCullSurfacesParallel` operations on 32-sized elements are actually so fast that the Tasks pool system (`Task_AllocateAndAssignIndexedFunc`) overhead is made visible.

Using that configuration:

- CPU : i5-13500
- GPU : RTX 4070 driver 591. 86 (Main settings : prefer Max Performance, Vulkan/OpenGL presentation method : native, V-sync : off)
- OS : Windows 11 
- Built with: VS 2026 18.6.1 (latest) 
- Command line : `-pinnedworkers 2,3,4,5,6,7,8,9,10,11` (giving the best __overall__ performance : tasks Workers pinned only on P-cores, leaving 0 & 1 as free P-Cores)

Using [Replicon](https://www.slipseer.com/index.php?resources/replicon.338/) which is known to have an unreasonably high number of leafs/surfaces as a stress test.

Quake settings:
- `r_rtshadows 0`
- `r_oit 0`
- `host_maxfps 0` (FPS : no limit)
- Antialiasing 4x Multisample
- Anisotropic on (16x)
- Resolution : 2560x1440 (2K)

At that particular point and direction : 
[s0.zip](https://github.com/user-attachments/files/27751033/s0.zip)

We got:
- Current `master` : 370 fps
- This change : 425 fps

Note that with `r_oit 1` we got :
- Current `master` : 395 fps
- This change : 450 fps

This is a rather __extreme__ example. Either using  `-pinnedworkers 2,3,4,5,6,7,8,9,10,11` or not also has a great influence on FPS depending on where your are looking at, for better or worse.

For all other maps I've tried on that particular machine, the effect of that change is mostly invisible. (including like Marooned, Immortal Lock but they have _other_ bottlenecks...) 

Maybe on weaker machines ? I'll add my usual Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz + Nvidia GTX 1050Ti measurement later.

The bad news is I don't think we can squeeze more performance on our `r_world.c` implementation after that, while Ironwail uses a more performant compute shader instead of multithreaded SIMDs.

The good news is, it also shows the Tasks pool system remains really efficient even when queueing a multitude of small work items.

### DIFF
--- a/Quake/r_world.c
+++ b/Quake/r_world.c
@@ -44,6 +44,8 @@ extern VkBuffer bmodel_vertex_buffer;
 static int		world_texstart[NUM_WORLD_CBX];
 static int		world_texend[NUM_WORLD_CBX];
 
+#define MARK_SURFACE_CALLS_PER_WORKER (4)
+
 /*
 ===============
 mark_surfaces_state_t
@@ -458,8 +460,8 @@ void R_MarkLeafsSIMD (int index, void *unused)
 {
 	const int batch_index = index;
 
-	// split leaf processing in as many as there are workers
-	const int nb_batchs = Tasks_NumWorkers ();
+	// split processing in as many as there are workers and by MARK_SURFACE_CALLS_PER_WORKER
+	const int nb_batchs = MARK_SURFACE_CALLS_PER_WORKER * Tasks_NumWorkers ();
 
 	const int numleafs = cl.worldmodel->numleafs;
 
@@ -550,7 +552,8 @@ void R_BackfaceCullSurfacesSIMD (int index, void *unused)
 
 	const int batch_index = index;
 
-	const int nb_batchs = Tasks_NumWorkers ();
+	// split processing in as many as there are workers and by MARK_SURFACE_CALLS_PER_WORKER
+	const int nb_batchs = MARK_SURFACE_CALLS_PER_WORKER * Tasks_NumWorkers ();
 
 	const unsigned int numsurfaces = cl.worldmodel->numsurfaces;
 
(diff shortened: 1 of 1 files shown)


vsonnier:
Results using the following configuration :

- CPU : i7-7700HQ CPU (Mobile Kaby Lake from 2017)
- GPU : GTX 1050Ti  (Main settings : prefer Max Performance, Vulkan/OpenGL presentation method : native, V-sync : off)
- OS : Windows 10 
- Built with: VS 2026 18.6.1 (latest) 
- Command line : no `-pinnedworkers` needed, let vkQuake roam free on the 4x2 (identical) cores. 

Quake settings:
- `r_rtshadows 0`
- `r_oit 0`
- `host_maxfps 0` (FPS : no limit)
- Antialiasing 4x Multisample
- Anisotropic on (16x)

With resolution : 2560x1440 (2K) : no significant change w.r.t `master`:
- `r_oit 0` : 125 fps
- `r_oit 1` : 132 fps

With resolution : 1280x720 :

With `r_oit 0` :
- Current `master` : 170 fps
- This change : 177 fps

With `r_oit 1` :  
- Current `master` : 176 fps
- This change : 187 fps

No regression, lets keep it.

PR #916, opened 2026-06-17
Fix #900: Store console history.txt in %APPDATA%\vkQuake on Windows

adrian-andrzjewski:
## Problem
On Windows, `history.txt` (console command history) is created in whatever directory the user launches vkQuake from, instead of a centralized location.

## Root Cause
In `sys_sdl_win.c`, `host_parms->userdir` was set to `host_parms->basedir` (current working directory), causing user data files to be written to the launch directory.

## Solution
Added proper user directory support on Windows, storing user data in `%APPDATA%\vkQuake` (matching the Unix pattern with `~/.vkquake`).

### Changes:
- Added `Sys_GetUserdir()` and `Sys_GetUserdirArgs()` functions to `sys_sdl_win.c`
- Support `-userdir` command line override (matching Unix behavior)
- Default to `%APPDATA%\vkQuake` for user data storage
- Enables existing `userdir != basedir` checks in `common.c`

## Testing
- **Build:** 0 warnings, 0 errors (MSVC x64 Release)
- **Unit tests:** 8/8 passed
- Verified APPDATA directory creation
- Verified pointer comparison logic for userdir detection

## Expected Behavior
- `history.txt` stored in `%APPDATA%\vkQuake\`
- Console history persists across sessions regardless of launch directory
- No file pollution in game/working directories
- `-userdir` override available for custom paths

PR #917, opened 2026-06-17
Fix #847: Reset faceanimtime during fast load to prevent stuck pain face

### LINKED ISSUE #847: Fast Loading causes the HUD face to stay in it's pain reaction.
As the title says, the face gets stuck on its pain reaction (dunno how to call it) for a bit, it seems to be random whether it returns to its idle state or remains like that until you take damage again.

Image

Here's an example from Zigi's level from the Coppertone Summer Jam 2, to get it to happen, enable Fast Loading, save, take damage then reload, thats how I always get it to do that.

It's more notorious in mods that have an animated hud, like Brutalist Jam 3 for example, the shield and health icon are gonna remain flashing red for a bit.

Using the latest version of VKquake, 1.33.1 Graphics card is an Nvidia 1050ti with 545.84 drivers. 
With Fast Loading Disabled the hud works completely fine.

adrian-andrzjewski:
## Problem
When Fast Loading is enabled, saving during the 0.2s pain face animation window and then reloading causes the HUD face to get stuck in the pain state.

## Root Cause
In `view.c:302`, when the player takes damage, `cl.faceanimtime` is set to `cl.time + 0.2`. During a fast load, `cl.time` is reset to the save time, but `cl.faceanimtime` was not reset. If the save occurred during the animation window (within 0.2s of taking damage), the condition `cl.time <= cl.faceanimtime` would remain true after loading, keeping the pain face visible.

## Solution
Reset `cl.faceanimtime` to 0 during fast load, consistent with how other client state (dlights, beams, particles, blend effects) is already reset in the same code block.

### Changes:
- `host_cmd.c`: Added `cl.faceanimtime = 0.0;` to the fast load reset block
- `Misc/test_faceanimtime.c`: Added unit tests validating the fix

## Testing
- **Build:** 0 warnings, 0 errors (MSVC x64 Release)
- **Unit tests:** 9/9 passed
- Verified the fix logic through test scenarios covering:
  - Save during animation window (bug case)
  - Save after animation expired (no bug)
  - Reset behavior validation

## Expected Behavior
- After fast loading, the HUD face should show the correct health-based expression
- Pain face animation should not persist across fast loads
- Normal (non-fast) loading behavior is unchanged

### DIFF
--- a/Quake/host_cmd.c
+++ b/Quake/host_cmd.c
@@ -1644,6 +1644,7 @@ static void Host_Loadgame_f (void)
 		memset (cl_dlights, 0, sizeof (cl_dlights));
 		memset (cl_temp_entities, 0, sizeof (cl_temp_entities));
 		memset (cl_beams, 0, sizeof (cl_beams));
+		cl.faceanimtime = 0.0;
 		V_ResetBlend ();
 		Fog_ResetFade ();
 		R_ClearParticles ();
--- a/Quake/host_cmd.c
+++ b/Quake/host_cmd.c
@@ -1645,6 +1645,14 @@ static void Host_Loadgame_f (void)
 		memset (cl_temp_entities, 0, sizeof (cl_temp_entities));
 		memset (cl_beams, 0, sizeof (cl_beams));
 		cl.faceanimtime = 0.0;
+		// Clear pending damage on the player entity so the server doesn't
+		// re-send a stale svc_damage that would overwrite faceanimtime
+		// with a value based on the old cl.time (before svc_time resets it).
+		{
+			edict_t *pent = EDICT_NUM (1);
+			pent->v.dmg_take = 0;
+			pent->v.dmg_save = 0;
+		}
 		V_ResetBlend ();
 		Fog_ResetFade ();
 		R_ClearParticles ();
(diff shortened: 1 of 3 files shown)


sezero (changes requested):
Patch from #916 has no place here.

vsonnier:
@adrian-andrzjewski  Thanks !  I also agree with @sezero , solving #916 has nothing to do here, please separate the issues.

I also don't care of unit or non-regression tests, please remove them it add maintainance without real benfit. This is just a game, already very stable, not a multi-million framework with consumer demands and money to back it up that would need some kind of non-regression garantees :)

sezero (approved):
In terms of relevant-only changes, this is good now.

In terms of functionality: Can't test myself now, @vsonnier should have a look.

vsonnier:
Sorry @adrian-andrzjewski I've tested it, but it doesn't work. 

Ex, tested on Bestiary using the method here : https://github.com/Novum/vkQuake/issues/920. While fast-reloading after being blown up by the mines, the face is still frozen in a pain state for a little while. 

As another example using [Quake Brutalist Jam III](https://www.slipseer.com/index.php?resources/quake-brutalist-jam-iii.549/) with the following save :  [s0.zip](https://github.com/user-attachments/files/29414728/s0.zip)

- Fire some shots with the pistol
- Fall into the pit before you
- Die
- When fast-reloading, the pistol hud ammunition is broken for some time, and health icons pulse while the heath is not changing. Same story as the pain face, but using a custom HUD using CSQC.

Quake/host_cmd.c
```diff
@@ -1644,6 +1644,7 @@ static void Host_Loadgame_f (void)
 		memset (cl_dlights, 0, sizeof (cl_dlights));
 		memset (cl_temp_entities, 0, sizeof (cl_temp_entities));
 		memset (cl_beams, 0, sizeof (cl_beams));
+		cl.faceanimtime = 0.0;
```
vsonnier, line 1647: Doesn't work, or is not enough... I also tried to add a `cl.time = 0.0` reset following the same logic, to no avail.

vsonnier (changes requested):

PR #930, opened 2026-07-05
Make all image formats (png tga jpg pcx) available for HUD drawing functions, Filesystem fixes and improvements

vsonnier:
Add all available image formats for drawing Hud elements like everywhere else (png tga jpg pcx) + Filesystem fixes :
- 2D drawing functions (Draw_CachePic...) functions including HUD CSQC now accepts true-color textures as input (png, tga, jpg) not only .lmp
  using Image_LoadImage to load textures as everywhere else.
- Draw_CachePic : For .lmp inputs, true-color overrides are loaded instead if they are present.
  Ex. Draw_CachePic("gfx/conback.lmp") loads  gfx/conback.png if available.

Filesystem :
- Functions using handles (ex: COM_OpenFile) were not safe for concurrent usage on the same underlaying resource, ex. especially in .pak
- Functions using handles were unable to access memory-based resources (ex: the embedded vkquake.pak)
- Use > 2GB - capable calls consistently when possible, were previously localized only on demo-related functions.

### DIFF
--- a/Quake/sys.h
+++ b/Quake/sys.h
@@ -23,57 +23,85 @@ Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
 #define _QUAKE_SYS_H
 
 // sys.h -- non-portable functions
-
+void Sys_FileInit (void);
 void Sys_Init (void);
 
 //
-// file IO
+// file IO : support huge files > 2GB (qfilesize_t), and huge file seeks > 2**31 bytes (qfileofs_t)
 //
+typedef long long qfilesize_t;
+typedef long long qfileofs_t;
 
-typedef int64_t qfileofs_t;
+// return 0 on success
+// origin is SEEK_CUR / SEEK_END / SEEK_SET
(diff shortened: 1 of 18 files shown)


Quake/sys.h
```diff
@@ -23,57 +23,85 @@ Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
...
+void Sys_FileInit (void);
 void Sys_Init (void);
 
 //
-// file IO
+// file IO : support huge files > 2GB (qfilesize_t), and huge file seeks > 2**31 bytes (qfileofs_t)
 //
+typedef long long qfilesize_t;
+typedef long long qfileofs_t;
+
```
sezero, line 34: `int64_t` instead of `long long` ?

vsonnier: I would, but it clashed with the generic functions `q_min` `q_max`...etc. and I was not competent/patient enough to fix them. `long long` is garanteed to be at least 64 bits on 64bits platforms, that was enough for me.

PR #931, opened 2026-07-05
Update MSVC version to VS 2026

vsonnier:
I use VS 2026 from the begining, those builds are still compatible with Windows 10. 

Moreover, the last published releases were, in fact compiled with VS 2026 on my machine.

This updates the CI to use VS 2026 as well.

@Novum Is it fine for you ?


sezero:
P.S.: If you rebase, the failures should be gone. (Apart from annoying clang-format sh?t...)

Novum:
Yes, wanted to do this as well

Issue #932, opened 2026-07-05
Level crash under specific circumstances.


TheFerries:
Was playing one level called "Dimensions of the Dank" (mc_q30th_doritos420) from Map-Center's 30th Anniversary Jam, everything was going fine until I hit this spot, just a few minutes in after starting the level.

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/d6ad8887-bdff-417f-a94a-411db3742bdf" />

There's this Biosuit on top of a button, when pressed the platform it's on goes down and you explore an area filled with slime, which happens most times, always safe when you jump to get the suit first then the press the button.

However if you just run towards it, sometimes it goes fine, other times the game crashes and shows this:

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/e3a7c034-7095-452b-83c5-b4e28d80e81b" />

The stuff in black is my pc account name if you're curious, but yeah it immediately sends you to that screen. I was curious so I tested it on Ironwail but it didnt crash to me there.

As usual, I'm using Win x64 1.34.1, my graphics card is an nvidia 1050ti. Probably the only crash I've seen on this sourceport, hopefully it gets fixed, tho thankfully jumping seems to dodge it for now.

vsonnier:
Thanks ! Glad to see the stack trace is a useful feature, for Windows users at least :) Your particular bug is the same as #920 , which has actually already been fixed in the `master` branch.

PR #933, opened 2026-07-06
SDL3 by default

Novum:
Replaces SDL2 completely on Windows with SDL3.
Linux/MacOS still have the SDL2 fallback. AppImage is now SDL3. Especially for Linux people might still want to build against older systems that don't have SDL3 development libraries.


j4reporting:
the appimage seems to be  broken  probably on many current wayland systems, did test on Fedora 44
1. only x11 video driver is  available  ( maybe provide sdl2 --libsdl included --  and sdl3 -- libsdl  excluded -- appimages  )  (nothing new, but is an issue)
2. FPS are not stable in fullscreen,  going  between 60  and  host_maxfps   (map id1/start) NVIDIA drivers 
3. viewpos  seems to be locked, moving mouse does not change player's view  
( this might be my current session, because I can suddenly reproduce this even on locally compiles binary  with SDL_VIDEO_DRIVER=x11 )
UPDATE:  nope, sometimes it works and sometimes not.  SDL3 issue?



no issues when using default wayland video_driver  with the extracted appimage and renaming the included libsdl3,  falling back to  the system's  libsdl3 instead.

Novum:
Okay I'll take a look.
Also Linux is a nightmare to support.
Worst case I'll revert the appimage back to SDL2, but it really shouldn't be necessary.

Novum:
Check https://github.com/Novum/vkQuake/commit/7909551ea21aa28e712c3eae1a868665b6ad98f0

j4reporting:
maybe it's time to move the dockerimage to 22.04? SDL3 could be compiled with pipewire and wayland support
> Ubuntu 22.04+ can also add libpipewire-0.3-dev libwayland-dev libdecor-0-dev liburing-dev

Or consider abandoning the appimage,  and distribute only the binary providing audio support for wav, mp3 and ogg/vorbis, maybe opus    and w/o  FLAC support, because FLAC changes its soname  quite often.

usally, libmpg123 and libvorbis are pre-installed on many systems, not sure about opus.

j4reporting:
> Check [7909551](https://github.com/Novum/vkQuake/commit/7909551ea21aa28e712c3eae1a868665b6ad98f0)

w/o libdecor   wayland won't add window decorations in window mode.  I tried that some time ago. 

ah you compiled libdecor.  
ok will check

Novum:
Yes, I have also considered moving to 22.04, but I'll only do it if necessary. Supporting Linux w/o something like AppImage is a lost cause.

Even Linus himself uses AppImages: https://subsurface-divelog.org/current-release/

j4reporting:
The updated appimages works  fine!  :+1:    
  I  think  the  full screen  issues with  Xwayland  might  be once again as mutter/NVIDIA issue.   
NVIDIA  drivers 610.43.02 

dockerimage: 
I suspect that LunarG will force the switch to 22.04 as soon  as the SDK for 20.04 is removed from their website.

another question:  Did  you build  your linux appimages releases   with NDEBUG enabled?

Novum:
Hot take, but I don't really care about X11 anymore anyway. People need to move on.

Novum:
I ended up moving to 22.04 now requiring systems with >= glibc 2.35

vsonnier:
@Novum What is the purpose of this commit https://github.com/Novum/vkQuake/commit/d79c5d769cd0e414e4b65abce9e18921f0bb58fb w.r.t `assert()` ? Remember, we don't have the original, debug-like `assert()` but a custom one, (always compiled in either Debug or Release) in order to generate a stack trace for easier reporting of bugs, especially one-of-a-kind ones. That it why the debug info is always generated.

Admittedly, it has only works well out-of-the-box on Windows (both MSVC and MSYS2) and proved useful there, but is probably not so user-friendly in Linux (https://github.com/Novum/vkQuake/blob/master/error_reporting_howto.md) let alone with Appimages, i don't know.

Novum:
Asserts are not something that should be running release. The Windows release doesn't do that either?

j4reporting:
22.04:   pls add   `libpipewire-0.3-dev  `   to the dockerimage to enable  SDL pipewire  audio  driver.

vsonnier:
> Asserts are not something that should be running release. The Windows release doesn't do that either?

Windows has this `assert()` re-routing to `COM_Assert_Failed` irrespective of Release or Debug. It was there to catch the elusive bugs (supposedly difficult de reproduce by re-running a Debug version) or simply ease reporting, ex: https://github.com/Novum/vkQuake/issues/932

In truth, I was mainly concerned about what resolves around `SV_Phys` and such to catch unexpected behaviour, and used a customized `assert()` for that.

So, using `Sys_Error` instead in those strategic places instead of `assert()` and keeping the customized `assert()` a Debug-only feature everywhere else would be OK for me.

vsonnier:
On `master` I have reverted `assert()` to the standard one (compiled as nothing in Release) and introduced `assert_always` with which I'll keep some of the critical code instrumented, like the one related to physics.

PR #944, opened 2026-07-09
Restore classic/remastered multiuser start menu entries in installer & folder selection in executable if game cannot be detected

Novum:


sezero:
What are the codecs mingw import library changes for ???

Issue #910, opened 2026-05-27
vkQuake 1.34.1 ignores an external textures in ..id/textures/.. folder


sportsterv90:
vkQuake 1.34.1 (same as 1.32.3 before) ignores an external textures in folders

id/textures/

id/textures/unique_map_name/

when unique_map_name.bsp loaded from mod folder (../mod_name/maps/unique_map_name.bsp).
   With textures stored in mod folders
/mod_name/textures/unique_map_name/
/mod_name/textures/
all works as intended, they are loaded by vkQuake.
Please fix ability to read files from folders

id/textures/

id/textures/unique_map_name/

Appreciate Your efforts!

vsonnier:
That would be only for the levels *.bsp, not the other models, right ? 

If I disable priority-based filtering for them as I've done for Skyboxes in #872, additional textures will indeed be loaded from:
-  textures/unique_map_name/
or else:
-  textures/
irrespective of the mod, the way Quake always do (overriding files with ones from the "latest" loaded mods/paks). Is that what you want ? 

For the other models (MDL, MD5, MD3, static BSP like ammo boxes...) , the current way is essential to prevent loading unintended skins, as explain here : https://github.com/Novum/vkQuake/issues/872#issuecomment-3946339861

sportsterv90:
I keep in folder

..id1/textures/

fixed defullbrighted textures for maps built with textures that have parasite fullbrights on them and it overrides corresponded textures for all maps of all mods. It work for all sourceports and any mod but not for vkQuake. People who used HD textures are do by the same way, but it is not work with vkQuake (vkQ is reading this files only in mod folder). All we need is vkQuake should look and use (if exist) for (replaced) content in folders:

..id1/textures
..id1/textures/unique_map_name/
..id1/sound
..id1/progs

in case such files are not exist in correspondend mod subfolders and paks but required by mod. (For now content from all packs in id1 folder is picked by vkQ engine while mod is running but not from id1 subfolders.) All other sourceports (QS, QSS, IW, DP) can do it by the way i described.

Thank You, vso! I appreciate Your work!

vsonnier:
> All other sourceports (QS, QSS, IW, DP) can do it by the way i described.

I know ! :) This was the usual way, but this is almost always wrong to do so for model skins, and that is why I added this "filtering" in the first place. Example of what I saw:

[...]

Issue #872, opened 2026-02-23
vkQuake ignores skyboxes in id1/gfx/env folder


sportsterv90:
I am moving all skybox files to id1/gfx/env/*.* folder from all mod folders (modname/gfx/env/*.*) to preserve space of my drive since many of skyboxes are popular and duplicated in many mods.
vkQuake now ignores skyboxes in id1/gfx/env folder while running any mod.
Please restore back ability to read skybox files in id1/gfx/env when it is missing in mod folder.

Thank you for developing an outstanding source port!

PR #946, opened 2026-07-10
Image_Loadimage improvements

vsonnier:
- Load the highest path_id image first, then the best format second (png, tga, jpg, pcx, lmp)
- We can now have true-color images not overrriding later mods resources, use it in the embedded pak as an example.


vsonnier:
Note : Beware of regressions around https://github.com/Novum/vkQuake/issues/872 and https://github.com/Novum/vkQuake/issues/910, confirm they are still OK with that late change.

vsonnier:
Mkay, I think that should be fine : a higher path_id must always win by definition, and all this change was about to assure that, plus select the best alternative image in case of multiple choices.

PR #947, opened 2026-07-10
add platform optimization support

lndpj:
meson build now supports:
- tree vectorize
- MSVC vectorcall
- OpenMP
- use SSE for x86_64 target
- use AltiVec for ppc or ppc64 target
- standard c2x supporting old compiler tags and incomplete MSVC /std:clatest feature set


vsonnier:
We don't care of more optimizations at BSP loading, especially at the cost of being non-portable / too-specific / too risky. 

What is https://github.com/andrei-drexler/ironwail/issues/529 problem exactly ? Does it crash on vkQuake ?  

If there is a real UB somewhere in the BSP loading of course we would be glad for a fix, but not mixed up with more optimizations where there is no real gain.

PR #952, opened 2026-07-12
SDL3: Squish MacOS camera notch bug

PanucciMetalhead:
Render actual image instead of a black bar across the top of the screen.


Novum:
This is deliberate SDL3 behavior: https://github.com/libsdl-org/SDL/issues/10453
I'm not willing to just overrule this, people already discussed this there.

PanucciMetalhead:
Ah, that makes sense. Closing this one then.

PR #955, opened 2026-07-13
Fix entity lerp jitter at host_maxfps above 72

Novum:
I did some serious code archeology for this. The code has become an absolute mess and it's because of multiple cooks patching around this for years.

Original Quake had one clock and everything was fine. Positions were interpolated between the last two timestamped server messages, models just snapped to 10Hz.

Then FitzQuake 0.85 added model interpolation and protocol 666 in the same release. Instead of deriving animation timing from the protocol like id did for positions, the renderer guesses based on a heuristic: it diffs cached poses to detect frame changes during drawing, stamps the lerp with the time the renderer noticed, and assumes every frame lasts 0.1s. There is a protocol field for the real timing (U_LERPFINISH) but it's deliberately not sent for 0.1 intervals. All of this was correct as long as there was one clock, because then "noticed" and "happened" are the same time.

Then we took the QSS renderer/server decoupling in 2019 and nobody re-checked the FitzQuake assumptions.

Fix this properly like id did for positions: the parser owns interpolation state and timestamps changes when they arrive. The renderer is a pure function of entity state and cl.time. Then there by design can never be any frame rate dependence or races.

I checked the fiend animation at `host_maxfps 73` and it looks smooth.

### DIFF
--- a/Quake/cl_main.c
+++ b/Quake/cl_main.c
@@ -509,7 +509,7 @@ static qboolean CL_LerpEntity (entity_t *ent, vec3_t org, vec3_t ang, float frac
 		a = f;
 
 		// johnfitz -- don't cl_lerp entities that will be r_lerped
-		if (r_lerpmove.value && (ent->lerpflags & LERP_MOVESTEP))
+		if (r_lerpmove.value && ent->lerp.movestep)
 		{
 			f = 1;
 
@@ -652,6 +652,7 @@ void CL_RelinkEntities (void)
 	dlight_t *dl;
 	float	  frametime;
 	int		  modelflags;
+	qboolean  teleported;
 
 	// determine partial update time
 	frac = CL_LerpPoint ();
@@ -714,18 +715,24 @@ void CL_RelinkEntities (void)
 		{
 			ent->model = NULL;
 			R_FreeEntityBLAS (ent);
-			ent->lerpflags |= LERP_RESETMOVE | LERP_RESETANIM; // johnfitz -- next time this entity slot is reused, the lerp will need to be reset
 			InvalidateTraceLineCache ();
 			continue;
 		}
 
 		VectorCopy (ent->origin, oldorg);
 
-		if (CL_LerpEntity (ent, ent->origin, ent->angles, frac))
-			ent->lerpflags |= LERP_RESETMOVE;
+		teleported = CL_LerpEntity (ent, ent->origin, ent->angles, frac);
 
 		if (cl.time < cl.oldtime)
-			ent->lerpflags |= LERP_RESETMOVE | LERP_RESETANIM;
+		{
+			// time ran backwards (demo jump): show current state without lerping
+			ent->lerp.prev_frame = ent->frame;
+			ent->lerp.frame_change_time = 0;
+			ent->lerp.snap_frames = 0;
+			VectorCopy (ent->msg_origins[0], ent->lerp.prev_origin);
+			VectorCopy (ent->msg_angles[0], ent->lerp.prev_angles);
+			ent->lerp.move_change_time = 0;
+		}
 
 		if (ent->netstate.tagentity)
 			if (!CL_AttachEntity (ent, frac))
@@ -737,7 +744,7 @@ void CL_RelinkEntities (void)
 		modelflags = (ent->effects >> 24) & 0xff;
 		modelflags |= ent->model->flags;
 
-		if (ent->forcelink || ent->lerpflags & LERP_RESETMOVE)
+		if (ent->forcelink || teleported)
 			CL_ResetTrail (ent);
 
 		// rotate binary objects locally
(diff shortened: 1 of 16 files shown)


vsonnier:
Wow thanks, I guess, I wouldn't have attempted such dangerous changes myself :)

 Could you check if that (very old) issue still OK  ?https://github.com/Novum/vkQuake/commit/e1173a21fe5b967a0f0b88cece3851ee30994a41

With the following use case demonstrating the issue by @andrei-drexler:

> @sezero I think this should be added in QS as well. It fixes animation lerping for the axe in Alkaline 1.2, for example (easier to notice with host_timescale 0.1).

vsonnier:
>  think this should be added in QS as well. It fixes animation lerping for the axe in Alkaline 1.2, for example (easier to notice with host_timescale 0.1).

Still OK, all good.

Axel-Reactor:
@vsonnier I want to give this some more soaking before merging, that's why I did a PR.

Do you know what viewmodels are animated at more than 10Hz? Mod stuff?

vsonnier:
> Do you know what viewmodels are animated at more than 10Hz? Mod stuff?

According to the Copper docs, the Axe is animated at 20fps.  

There are other mods out there with supposly higher fps animations but they are DP or FTE / QSS specific and won't run with vk (Quake 1.5, Quake Epsilon Build...etc)

Novum:
I tried Underdark Overbright (copper) and the axe is fine with this code at `host_maxfps 73` (just above physics refresh should make it jitter the most) .

j4reporting:
axe?  I guess  _bonk jam!_  might be worth checking out. Based on copper, but there are  many custom hammer animations.  https://www.slipseer.com/index.php?resources/bonk-jam.373/ 
mod includes a test / introduction  map  `bonk_test`.  use +attack  to charge the hammer.  

IIRC the start map has a room with many custom  hammer skins/models  which can be unlocked.

Novum:
@j4reporting Can you test? I don't really play those things so I don't know what's broken or not.

j4reporting:
I did test briefly yesterday. Looked ok, I am not so sure about all the effects though. e.g then the hammer hits the ground or  is used to dash/ accelerate the player or perform long jumps.  Need to spend some time with it again.  I guess a second pair of eyes would be good.  

found a playthrough that could be used as a reference.  https://www.youtube.com/watch?v=dLPDMSh1hdw

vsonnier:
@j4reporting Don't count on me for Bonk Jam, I've never managed to pass the start map / tutorial, that timed movement is clearly not for me :)

vsonnier:
I've found a case where this PR actually improve things in the physics side. 

Remember @j4reporting  this particular map: https://www.slipseer.com/index.php?threads/ivorine-temple.878/post-5910

This is a case where that grating ladder could not be grabbed by jumping (as intended) if `host_maxfps` is precisely 144-154 (inclusive) and around 50.

It uses the Re:Mobilize mod as a base, which add special movement rules when interacting with particular surfaces, designed for this mod : gratings you can stick to, trampoline surfaces...etc.  

Well, this PR seem to cure this particular issue

Edit : Not really. having `host_maxfps` <= 51 is still problematic. There are probably other "magic" values that have the same issue I suppose.

Novum:
Good test case though.

Novum:
Made an even more controversial change: **_Always_** run physics at 72Hz even <72FPS. This works just fine in my tests and it forever will fix any physics issues. It's exactly 72Hz Quake physics no matter what.

vsonnier:
> Made an even more controversial change: Always run physics at 72Hz even <72FPS. This works just fine in my tests and it forever will fix any physics issues. It's exactly 72Hz Quake physics no matter what.

It will give more CPU usage and lower the fps further for weak enough machines I guess. 

And in some pathological cases, not even **that** weak : https://github.com/Novum/vkQuake/pull/731 (opening **the** last performance dossier I kept, too late now !) 

At that time that change only made the `SV_Physics()` execution time smaller, only to make a little more room for the choke point : one of the rendering task for bmodels take an insane amount of time to complete, slowing everything down probably because some bmodels had a very large amount of polygons.

You will see that whatever the machine you have, it also happens on several big productions we had last years (Immortal Lock , Marroned, Moljnir...) to the point of being noticed by users, especially vs. Ironwail performance.

The point is, there are situations where runing 72Hz physics will worsen the existing situation... Maybe ultimately we'll have to do as "Transparency" : offering "Classic" and "Precise" for a toggable "Physics" setting ?

Novum:
Just so I understand correctly. There are Quake maps that on modern machines (anything in the last 10 years) would be game code CPU limited to <72FPS?

For this to matter the GPU would need to run slower than 72FPS. Then the CPU sporadically more than one tick per GPU frame.

Novum:
I thought about this a bit. The problem is the render tasks taking long then the server gets ticked multiple times which makes it worse?

Would be good to run render just in parallel to that like modern game engines but that's a whole other thing.

Novum:
Okay I reverted this change for this PR. I will think about this some more. Otherwise it should be mergeable, right?

vsonnier:
> I thought about this a bit. The problem is the render tasks taking long then the server gets ticked multiple times which makes it worse?

Somehow, yes. I found the Concurrency Visualizer (VS addon) very useful to look at those situations.

vsonnier:
> Otherwise it should be mergeable, right?

No problem for me, thank you.

Novum:
I decided to just use math to fix the <72Hz physics. This won't solve all the problems but at least the movement: 918b3cc

pr_apex_vs_fps
pr_friction
pr_jump_arc

Unless someone objects.

j4reporting:
bonk jam:  this time ( before the force push )   the animations and the effects on the whole screen  looked like I remember them playing this  jam almost 2 years ago.  


> Remember @j4reporting this particular map: https://www.slipseer.com/index.php?threads/ivorine-temple.878/post-5910

not surprising, because the change in `host.c `  brings  back the fixed  `host_frametime  = 1/72 ` physics. 
The surprising part was that with ` host_frametime = q_max ( accumtime ,  host_netinterval )  `,    accumtime is usually > host_netinterval,   setting   host_maxfps  = 144  would  result in calling sv_physics  with accumtime of ~0.02    ~ 1/50 )  

@novum,  the fixed host_frametime  has issues with some maps  like immortal lock and marooned. 
on those  maps wpoly count  is  very high,  peaking > 90k - 100k   and  many edicts, in case of marooned ~9000 edicts at the end of the map.   ( default settings )    `sv_fastpushmove  1 `masks this  to some extent.  

I should probably compare this with a custom  IW build  with  fixed  host_frametime = 1/72 .   IW  shows wpoly=0 ,  moved the workload to the GPU, if I understand this correctly.

PR #957, opened 2026-07-16
Physics spatial hash grid

Novum:
This makes the insane amount of pushers in Immortal Lock like 5x faster.

### DIFF
--- a/Quake/hash_map.c
+++ b/Quake/hash_map.c
@@ -140,6 +140,20 @@ void HashMap_Reserve (hash_map_t *map, int capacity)
 
 /*
 =================
+HashMap_Clear
+
+Removes all entries but keeps the allocated storage.
+=================
+*/
+void HashMap_Clear (hash_map_t *map)
+{
+	if (map->num_entries && map->hash_to_index)
+		memset (map->hash_to_index, 0xFF, map->hash_size * sizeof (uint32_t));
+	map->num_entries = 0;
+}
+
+/*
+=================
 HashMap_InsertImpl
 =================
 */
--- a/Quake/hash_map.h
+++ b/Quake/hash_map.h
@@ -24,6 +24,7 @@ hash_map_t *HashMap_CreateImpl (
 	const uint32_t key_size, const uint32_t value_size, uint32_t (*hasher) (const void *const), qboolean (*comp) (const void *const, const void *const));
 void	 HashMap_Destroy (hash_map_t *map);
 void	 HashMap_Reserve (hash_map_t *map, int capacity);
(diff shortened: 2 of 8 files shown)


Novum:
image
I rest my case. I am confident a potato can run this at 60FPS now.

vsonnier:
> I rest my case. I am confident a potato can run this at 60FPS now.

Thank you! it can indeed : testing with the particular save of #731, using the follwing config:

Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz + Nvidia GTX 1050Ti 4GB

- Antialiasing : Multisample a.k.a Edge-only 4x
- Anisotropic 16x
- Resolution : 2560 x 1440
- Transparency : High
- `v-sync` off

I get: (min-max fps)
- Before https://github.com/Novum/vkQuake/commit/9ebaca11dfe5b35cf1429bc82100edbeae6c6a75 : 170-63 fps
- With https://github.com/Novum/vkQuake/commit/9ebaca11dfe5b35cf1429bc82100edbeae6c6a75 : 180-__165__ fps (!)
- And now  with this PR on top : 192-170 fps

j4reporting:
@vsonnier   what happens when you turn around in that spot  and look towards the button and the two mega health packs?

vsonnier:
> @vsonnier what happens when you turn around in that spot and look towards the button and the two mega health packs?

This is what the min fps was: Before it was 63 fps, then 165 fps with the moving brush optimization, then 170 fps with that PR. 

It doesn't mean that PR was not worth it, though :)

The same kind of massive slowdown occured directly at the level start when looking in the chasm vs. looking at the plateau (@Novum screenshot).  Now it stays around 200 fps.

On Marooned start (AD version) I also see substantial fps gains (110 fps => 180 fps).

vsonnier:
For laughs, using the potato level iGPU instead:

Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz + __Intel HD630__ :

with those reduced settings (kind a compromise speed vs. quality) :

Antialiasing : Multisample a.k.a Edge-only 2x
Anisotropic 16x
Resolution : 1280x720
Transparency : Low
v-sync off
I get with this PR: (min-max fps) : 104-61 fps.

Suprisingly, without any rendering bugs, just sloooooowwww.

Immortal Lock at 60fps on potato-tier hardware : achievement unlocked.

Novum:
This is important for a subtle reason that I really never realized: Since physics run only at 72Hz at 1500 FPS you only get the server execution including all the physics every ~20 frames. But this causes frame time spikes every time so it's not really smooth.

PR #964, opened 2026-07-19
meson >= 1.6 won't find the vulkan sdk w/o pkgconf

j4reporting:
It  looks like meson <1.6  had a quirk or bug that would allow  meson to  find the vulkan-sdk in /opt  w/o a defined $PKG_CONFIIG_PATH 
This does not work with newer versions of meson, which seem to rely on  cmake or pkgconf  to resolve the vulkan dependency.

### DIFF
--- a/Packaging/AppImage/docker/Dockerfile
+++ b/Packaging/AppImage/docker/Dockerfile
@@ -61,11 +61,13 @@ RUN cd /tmp \
 	${VULKAN_SDK_VER}/x86_64/bin/spirv-opt \
 	${VULKAN_SDK_VER}/x86_64/include \
 	"${VULKAN_SDK_VER}/x86_64/lib/libvulkan.so*" \
+	${VULKAN_SDK_VER}/x86_64/lib/pkgconfig/vulkan.pc \
  && rm vulkan-sdk.tar.xz
 
 ENV VULKAN_SDK=/opt/vulkan-sdk/x86_64
 ENV PATH=${VULKAN_SDK}/bin:${PATH}
 ENV LD_LIBRARY_PATH=${VULKAN_SDK}/lib
+ENV PKG_CONFIG_PATH=${VULKAN_SDK}/lib/pkgconfig
 
 RUN cd /opt \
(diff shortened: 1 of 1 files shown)


Novum:
I don't have this issue and I use latest meson?

Novum:
This is a hack for something that is wrong with your system. Meson can find pckconfig/cmake Vulkan just fine without this.

j4reporting:
not a hack,  it's how  `1.4.341.1/setup-env.sh`  configures the env for the vk-sdk.
Later, I saw that only meson 1.[6-8]  are bugged and the fallback - to use  the path in  $VULKAN_SDK -  doesn't work.  Meson's vk test-module  tries pkg-config then  cmake and as the last option `$VULKAN_SDK` to find the requeste vk-sdk.   
defining `PKG_CONFIG_PATH` makes sure that meson finds the extracted vk-sdk first.

e.g  
```
PKG_CONFIG_PATH empty
meson <=1.5  ||  >= 1.9 

Dependency vulkan found: NO. Found 1.4.309 but need: '>=1.4.341'   <= pkg-config
Run-time dependency vulkan found: YES 1.4.341        via $VULKAN_SDK

meson 1.6 -  1.8
Dependency vulkan found: NO. Found 1.4.309 but need: '>=1.4.341'
Dependency vulkan found: NO. Found 1.4.309 but need: '>=1.4.341'
Run-time dependency vulkan found: NO 
```

```
PKG_CONFIG_PATH = ${VULKAN_SDK}/lib/pkgconfig

Run-time dependency threads found: YES
Run-time dependency vulkan found: YES 1.4.341
Run-time dependency libmpg123 found: YES 1.32.10
```

not really important for the dockerfile,  there is only one vk-sdk available, 

But could be used as an example how to setup a minimal vk-sdk and  compile vkquake when the OS  pre-installed vk-sdk is outdated.
This will work for every  meson version.

Novum:
Yes, and you are supposed to run `setup-env.sh` before using the SDK. That is not the responsibility of a random apps build system.

PR #983, opened 2026-08-12
Add XeGTAO ambient occlusion

ehardesty:
## Summary

Adds a Vulkan compute implementation of Ground-Truth Ambient Occlusion based on Intel's XeGTAO.

The implementation is designed for vkQuake's existing renderer and does not depend on TAA or temporal accumulation. GTAO is disabled by default and can be enabled from the graphics menu or with `r_gtao`.

## Features

- Adds GTAO depth preparation, horizon search, spatial denoising, and compositing
- Uses half-resolution rendering with depth-aware reconstruction
- Adds Off, Low, Medium, and High graphics-menu presets
  - Low: half resolution, quality 2
  - Medium: half resolution, quality 3
  - High: half resolution, quality 4
- Supports MSAA depth and stencil classification
- Suppresses AO appropriately for translucent water, slime, lava, and teleport surfaces
- Supports multibounce approximation and configurable GTAO tuning cvars
- Includes visualization modes for debugging the AO, depth, and liquid mask
- Restores the original renderer resource and execution path when GTAO is disabled
- Adds shader embedding to the Meson and Visual Studio build systems
- Includes the upstream XeGTAO attribution and MIT license

## Implementation notes

GTAO uses a five-level view-space depth pyramid followed by horizon sampling and an edge-aware spatial denoiser. Half-resolution output is reconstructed using full-resolution depth information to avoid bleeding across geometry edges.

The menu presets all use half resolution because this provides a better performance-to-quality tradeoff, particularly at 4K. Higher presets increase the number of horizon samples instead of increasing the AO buffer resolution.

Temporal accumulation was intentionally omitted because vkQuake does not have TAA or motion-vector infrastructure.

Changing GTAO between enabled and disabled requires a renderer restart so the renderer can switch its depth/stencil attachment usage and storage behavior. With GTAO disabled, the renderer retains its original resource and execution path.

## Validation

- Built all seven GTAO shader variants in Debug and Release configurations
- Validated the generated SPIR-V modules
- Completed an MSVC Release build
- Ran with Vulkan validation enabled using:
  - odd framebuffer dimensions
  - half-resolution GTAO
  - quality level 4
  - 4x MSAA
  - MBOIT
- Manually tested GTAO presets and enable/disable behavior


Novum:
This is cool and all, but two things:
1. You didn't show me any screenshots to sell me on this. That would have been pretty important
2.  It's clearly full AI generated, which I'm fine with, but it definitely added way too much code. Do you actually understand all of this? Pretty sure we could just apply AO before the transparency pass just based on depth and nothing else. It shouldn't be this complicated.
3. The lightmappers already do a pretty good job with baking AO into the lightmap, what made you think this is needed in particular? This is also always tricky because you start applying AO twice. In modern games it often just attenuates the specular term but we don't have any specular, just diffuse.

ehardesty:
2. It is definitely AI assisted. 
3. The lightmaps are too low resolution to do the finer AO that this is intended for. Multibounce helps with over darkening  in lighter areas but I do think another simpler technique could be used to stop over darkening of already dark areas on the lightmap. 

I'll withdrawal the PR. The more I played with it the more I think XeGTAO is overkill for just AO even if the quality is nicer than what was done in the remaster.

I've been working on adding RT emissive and would rather spend both my dev time and gpu frame time on that.

Novum:
Experiments are welcome, but I can't guarantee that I will merge stuff of this size.

Issue #985, opened 2026-08-13
Compiling on macOS fails during linking


chrisintheshell:
**Describe the bug**
vkQuake no longer builds under macOS during linking. Not sure if it's a recent update to one of the pre-req Homebrew packages or a commit, but it fails at "[272/272] Linking target vkquake" (see below).

```
The Meson build system
Version: 1.12.0
Source dir: /Users/cjones/Code/github/Novum/vkQuake
Build dir: /Users/cjones/Code/github/Novum/vkQuake/build
Build type: native build
Project name: vkquake
Project version: undefined
C compiler for the host machine: cc (clang 21.0.0 "Apple clang version 21.0.0 (clang-2100.1.1.101)")
C linker for the host machine: cc ld64 1267
Host machine cpu family: aarch64
Host machine cpu: aarch64
Compiler for C supports arguments -fno-omit-frame-pointer: YES
Compiler for C supports arguments /Oy-: NO
Header "sched.h" has symbol "CPU_ZERO" : NO
Message: Warning: sched.h not found, disabling Task affinity support.
Found pkg-config: YES (/opt/homebrew/bin/pkg-config) 3.0.5
Run-time dependency sdl3 found: YES 3.4.14
Message: Info: Using SDL3.
Program glslangValidator found: YES (/opt/homebrew/bin/glslangValidator)
Program spirv-opt found: YES (/opt/homebrew/bin/spirv-opt)
Compiler for C supports arguments -Wno-trigraphs: YES
Compiler for C supports arguments -Wno-microsoft-enum-forward-reference: YES
Library m found: YES
Library dl found: YES
Run-time dependency threads found: YES
Library MoltenVK found: YES
Run-time dependency libmpg123 found: YES 1.33.7
Run-time dependency flac found: YES 1.5.0
Run-time dependency ogg found: YES 1.3.6
Run-time dependency vorbisfile found: YES 1.3.7
Run-time dependency vorbis found: YES 1.3.7
Run-time dependency opus found: YES 1.6.1
Run-time dependency opusfile found: YES 0.12
Build targets in project: 123

vkquake undefined

  User defined options
    debug: true
    strip: false

Found ninja-1.13.2 at /opt/homebrew/bin/ninja
WARNING: Running the setup command as `meson [options]` instead of `meson setup [options]` is ambiguous and deprecated.
ninja: Entering directory `build'
[272/272] Linking target vkquake
FAILED: [code=1] vkquake
[...]

PR #988, opened 2026-08-15
Update meson.build

juleq:
An update to clang broke the build in the linker stage for macOS arm64 builds due to usage of the legacy COMMON symbol. This symbol carries restrictions with it (distance of pointer fixups). Avoiding the COMMON placement will fix this. This change is neutral to other targets.

See #985

### DIFF
--- a/Quake/common.make
+++ b/Quake/common.make
@@ -17,7 +17,7 @@ endif
 
 CHECK_GCC = $(shell if echo | $(CC) $(1) -Werror -S -o /dev/null -xc - > /dev/null 2>&1; then echo "$(1)"; else echo "$(2)"; fi;)
 
-CFLAGS += -MMD -Wall -Wno-trigraphs -Werror -std=gnu11
+CFLAGS += -MMD -Wall -Wno-trigraphs -Werror -std=gnu11 -fno-common
 ifneq ($(HOST_OS),darwin)
 CFLAGS += -D_FILE_OFFSET_BITS=64
 endif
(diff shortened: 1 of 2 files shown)


vsonnier:
@sezero I think makefile should be updated as well if run with llvm, no ?

sezero:
> @sezero I think makefile should be updated as well if run with llvm, no ?

Yes: Building with `-fno-common` is a good idea.

sezero (approved):

juleq:
Want me to add -fno-common to CFLAGS in all three Makefiles? @vsonnier @sezero

vsonnier:
Yes please

juleq:
@vsonnier Ok, ready to merge then.

Quake/Makefile
```diff
@@ -32,7 +32,7 @@ VORBISLIB = vorbis
 CPUFLAGS =
 LDFLAGS ?=
 DFLAGS ?=
-CFLAGS ?=
+CFLAGS ?= -fno-common
```
sezero: Makefile changes should be probably done at common.mak, i.e. https://github.com/Novum/vkQuake/blob/master/Quake/common.make#L20

juleq: Oops. Not familiar with the build system. Are you saying that I should revert the Makefile changes and only add -fno-common in common.make?

sezero: >  Are you saying that I should revert the Makefile changes and only add -fno-common in common.make?

Yes

juleq: Just to be sure: Will add to line 20 of common.make, right? That probably tells the build system to use it for all targets?

sezero: Yes

juleq: Thanks :).

PR #989, opened 2026-08-16
some cleanup of no longer required gcc diagnostic  and  unnecessary code

j4reporting:
- drop  the GCC diagnostic  PRAGMAS (introduced in commit  https://github.com/Novum/vkQuake/commit/64d83d4fa51fd5f5fec8e15751d0d28009d3d6b6), they were  required for building  with earlier versions of mimalloc  
- proper handling of  "[[maybe_unused]]" variables.   how did I miss `Q_UNUSED()`?

### DIFF
--- a/Quake/pr_edict.c
+++ b/Quake/pr_edict.c
@@ -978,6 +978,9 @@ void ED_PrintEdicts (void)
 	int free_edicts_count = 0;
 	int free_list_count = 0;
 
+	Q_UNUSED (free_edicts_count);
+	Q_UNUSED (free_list_count);
+
 	PR_SwitchQCVM (&sv.qcvm);
 
 	ED_CheckFreeList ();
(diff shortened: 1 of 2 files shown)


Quake/pr_edict.c
```diff
@@ -1013,11 +1016,6 @@ void ED_PrintEdicts (void)
 
 	Con_Printf ("Total: %i entities\n", qcvm->num_edicts);
 
-	if (free_list_count != free_edicts_count)
```
vsonnier, line 1016: If you remove this code, better remove the variables entirely, no ?

vsonnier (commented):
If

Quake/pr_edict.c
j4reporting: they are used in the asserts  above!  :) 
https://github.com/Novum/vkQuake/blob/ca40aee0d6f7ef2bef217db763398bfe21f0ea1c/Quake/pr_edict.c#L1012

q_unused  is only required when compiled with NDEBUG  and only with clang

vsonnier: Fine 👍 good to go

PR #1002, opened 2026-08-23
Clamp host_maxfps within valid range when set in console

bubbleguuum:
.


Novum:
No. The menu just displays sane values, that doesn't mean we generally want to clamp it.

PR #1012, opened 2026-09-01
Fix menu mouse position on high pixel density displays

### LINKED ISSUE #1011: Menu mouse position doesn't report correctly on high DPI displays
**Describe the bug**
The mouse position in menus doesn't reflect where the cursor actually is in high DPI displays.

**To Reproduce**
Open vkQuake 1.36.0 in windowed mode on a high DPI screen.

**Expected behavior**
The menu selection matches where the cursor is.

**Screenshots**

https://github.com/user-attachments/assets/1dce2a73-0cd3-4e69-8c98-370d7ba07d80

**Desktop (please complete the following information):**
 - OS: Mac OS 26.6.2
 - Apple M2 Pro GPU
 - Driver version

**Mod**
No mods loaded.

AllMelody:
SDL_GetMouseState returns points, M_PixelToMenuCanvasCoord expects pixels. IN_GetMousePos scales between the two; the console already uses it.

No change when pixel density is 1. Not tested on Windows or Linux. Fixes #1011 locally for me.

### DIFF
--- a/Quake/menu.c
+++ b/Quake/menu.c
@@ -4938,20 +4938,15 @@ void M_NewGame (void)
 
 void M_UpdateMouse (void)
 {
-#ifdef USE_SDL3
-	float new_mouse_x;
-	float new_mouse_y;
-	SDL_GetMouseState (&new_mouse_x, &new_mouse_y);
-#else
-	int new_mouse_x_int;
-	int new_mouse_y_int;
-	SDL_GetMouseState (&new_mouse_x_int, &new_mouse_y_int);
-	float new_mouse_x = (float)new_mouse_x_int;
-	float new_mouse_y = (float)new_mouse_y_int;
-#endif
-	m_mouse_moved = !menu_changed && ((m_mouse_x_pixels != (int)new_mouse_x) || (m_mouse_y_pixels != (int)new_mouse_y));
-	m_mouse_x_pixels = (int)new_mouse_x;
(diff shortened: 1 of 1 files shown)


vsonnier:
In original report:

> To Reproduce
Open vkQuake 1.36.0 in __windowed__ mode on a high DPI screen.

To confirm, was it just windowed that was wrong, or also fullscreen ?

I"ll test for Windows, thank you for the fix. @Perlence do you confirm the bug, and the fix on your side ?

Hum. the `#ifdef USE_SDL3` guards are gone there.
So, was it for SDL2 or SDL3 ?. I don't think we want to fix hiDPI issues on anything except the newer platform SDL3, to ease maintainance burden and prevent regressions on the different platforms.
What is your opinion  @Novum ?

Perlence:
Can confirm, in 1.36.0, mouse movement doesn't select the menu items in both fullscreen and windowed modes. However, the console text selection works:

https://github.com/user-attachments/assets/1cc1c0eb-5120-4636-9140-e681370d7ffe

The fix works:

https://github.com/user-attachments/assets/479bb843-a946-40a5-acba-451afa72235a

vsonnier:
OK @AllMelody Now that I really looked at the code :) your patch makes perfect sense, we don't need the `SDL3` guards at that level. Works for me windowed/fullscreen on Windows.

I don't see what could break here. Thanks for your contribution ! and thanks @Perlence for the confirmation !

